PR #59, opened 2020-02-27
fix(client): Fix ENV var prefix to be consistent with NR standards

jthurman42:
The engineering standard for Environment variable prefixes at New Relic is `NEW_RELIC_`. Update the API Key and Region variables to use this prefix instead (and remove `Personal` from the env var name for the API key)

### DIFF
--- a/internal/client/client.go
+++ b/internal/client/client.go
@@ -43,8 +43,8 @@ func CreateNRClient(cfg *config.Config, creds *credentials.Credentials) (*newrel
 
 // applyOverrides reads Profile info out of the Environment to override config
 func applyOverrides(p *credentials.Profile) *credentials.Profile {
-	envAPIKey := os.Getenv("NEWRELIC_PERSONAL_API_KEY")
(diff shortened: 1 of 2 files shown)


internal/client/client.go
```diff
@@ -43,8 +43,8 @@ func CreateNRClient(cfg *config.Config, creds *credentials.Credentials) (*newrel
 
 // applyOverrides reads Profile info out of the Environment to override config
 func applyOverrides(p *credentials.Profile) *credentials.Profile {
-	envAPIKey := os.Getenv("NEWRELIC_PERSONAL_API_KEY")
-	envRegion := os.Getenv("NEWRELIC_REGION")
+	envAPIKey := os.Getenv("NEW_RELIC_API_KEY")
+	envRegion := os.Getenv("NEW_RELIC_REGION")
```
sanderblue: The environment variable naming convention for the client and TF provider is `NEWRELIC_****`

PR #70, opened 2020-02-28
chore(lint): enable golangci-lint for further lint checking

zlesnr:


build/lint.mk
```diff
@@ -20,10 +22,11 @@ GOTOOLS += github.com/client9/misspell/cmd/misspell \
            github.com/gordonklaus/ineffassign \
            github.com/timakin/bodyclose \
            golang.org/x/lint/golint \
-           github.com/llorllale/go-gitlint/cmd/go-gitlint
+           github.com/llorllale/go-gitlint/cmd/go-gitlint \
+           github.com/golangci/golangci-lint/cmd/golangci-lint
```
jthurman42: We should be able to remove all the specific linters and their Make targets now that this is working again.

build/lint.mk
```diff
@@ -20,10 +22,11 @@ GOTOOLS += github.com/client9/misspell/cmd/misspell \
...
 
 
-lint: deps spell-check gofmt govet golint ineffassign gocyclo bodyclose lint-commit
+lint: deps spell-check gofmt govet golint ineffassign gocyclo bodyclose lint-commit golangci
```
jthurman42: Remove all these, make the golangci the `lint` target.

build/lint.mk
jthurman42: Also need to add golangci-lint to `/tools.go` to get it into go.mod.

jthurman42:
Golangci-lint is notorious for breaking on updates, let's use a config file for it (I recommend this one: https://github.com/newrelic/newrelic-client-go/blob/master/.golangci.yml )

Issue #89, opened 2020-03-02
User-Agent / Service Name are not set


jthurman42:
### Description
We are unable to determine that an API call is being run from the CLI, and what version of the CLI is being used.  This causes issues when trying to tie errors back to the specific requests / versions of the CLI.

### Current behavior
Command Line API calls return as the downstream library user-agent/service name (`newrelic-client-go`)

### Expected behavior
User-agent and ServiceName headers are appropriately set to the CLI name and version.

PR #91, opened 2020-03-02
fix(config): set user agent and service name, add version package

sanderblue:
Resolves: #89

### DIFF
--- a/build/test.mk
+++ b/build/test.mk
@@ -13,7 +13,7 @@ SRCDIR       ?= .
 GO_PKGS      ?= $(shell ${GO} list ./... | grep -v -e "/vendor/" -e "/example")
 FILES        ?= $(shell find ${SRCDIR} -type f | grep -v -e '.git/' -e '/vendor/')
 
-LDFLAGS_UNIT ?= '-X version.GitTag=$(PROJECT_VER_TAGGED)'
+LDFLAGS_UNIT ?= '-X github.com/newrelic/newrelic-cli/internal/version.GitTag=$(PROJECT_VER_TAGGED)'
 
 GOTOOLS += github.com/stretchr/testify/assert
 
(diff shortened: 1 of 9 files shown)


build/test.mk
```diff
@@ -25,7 +25,7 @@ test: test-only
 test-only: test-unit test-integration
 
 test-unit:
-	@echo "=== $(PROJECT_NAME) === [ test-unit        ]: running unit tests..."
+	@echo "=== $(PROJECT_NAME) ${PROJECT_VER_TAGGED} === [ test-unit        ]: running unit tests..."
```
jthurman42: Why add the version tag (and break the beautiful output format followed by the rest of the make system?

sanderblue: I wasnt done. I should've opened a draft PR lol

cmd/newrelic/main.go
```diff
@@ -8,13 +8,14 @@ import (
...
 	"github.com/newrelic/newrelic-cli/internal/entities"
+	"github.com/newrelic/newrelic-cli/internal/version"
 )
 
 var (
 	// AppName for this CMD
 	AppName = "newrelic-dev"
 	// Version of the CLI
-	Version = "dev"
+	Version = version.Version
```
jthurman42: We don't need to set another global here, just pass it to `Execute` on line 39, or better yet is to remove completely from `main.go` and put directly into `command.go` inside of `Command`, and fix `versionCmd` to reference as well.

internal/client/client.go
```diff
@@ -15,6 +16,7 @@ func CreateNRClient(cfg *config.Config, creds *credentials.Credentials) (*newrel
 		err            error
 		personalAPIKey string
 		region         string
+		serviceName    = "newrelic-cli"
```
jthurman42: This should be a constant in the package, easier to find / change later.

internal/version/version.go
```diff
@@ -0,0 +1,4 @@
+package version
+
+// Version of the CLI
+const Version string = "0.1.0"
```
jthurman42: Add the Release makefile / scripts to automatically update this ala https://github.com/newrelic/newrelic-client-go/blob/master/build/release.mk

sanderblue: This felt out of scope for this PR, but I went ahead and did it.

PR #93, opened 2020-03-03
Build Docker container

jthurman42:
* Update the Container name to `newrelic/cli`
* Update CircleCI config for building docker / pushing the container

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -3,18 +3,109 @@ version: 2.1
 orbs:
   snyk: snyk/snyk@0.0.8
 
-jobs:
-  build:
-    docker:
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -3,18 +3,103 @@ version: 2.1
...
+            - test
+            - security-scan
+      - hold-for-deploy:
+          type: approval
+          filters:
+            branches:
+              only: master
+      - docker-push:
+          filters:
+            branches:
```
zlesnr: Is there a tags filter?  Do we want a new build/push for each commit, or just for releases?

jthurman42: Yup, looking to change that now.

PR #98, opened 2020-03-04
Automated Releases via goreleaser

jthurman42:
This redoes some of the version work to only use LDFLAGS so we can build assets with goreleaser.

* Uses goreleaser for creating artifacts
* Automated GitHub Releases
 * Builds Darwin/Linux/WIndows binaries (Attached to GH release)
* Builds and pushes docker images (three tags: latest, v0.2, v0.2.1)
 
New workflow:
* When ready to release, run `make release` to start the normal release build & PR process
* Merge release branch
* Create new version tag & push
* CircleCI should automatically create the release and push all artifacts up.


.goreleaser.yml
```diff
@@ -0,0 +1,69 @@
...
+    env:
+      - CGO_ENABLED=0
+    goos:
+      - linux
+      - darwin
+      - windows
+    goarch:
+      - amd64
+    ldflags:
+      - -s -w -X main.version={{.Version}} -X main.AppName={{.Binary}}
```
zlesnr: Are the flags here intended to be on the same line?

jthurman42: They do not have to be.  I can move it up, but makes for a long line and this works correctly #yaml

zlesnr: Two lines is fine, just surprised that it works.

jthurman42: YAML is weird, but it's the indention that makes it work...

PR #102, opened 2020-03-04
tests(config): fix tests that were overriding actual config files

### LINKED ISSUE #99: Config integration tests overwrites user config.yml
### Description
When running tests locally / local dev environment, the config integration tests is overwriting the user config.  Looking at the test, it appears to specify a directory to use instead, but it appears not to be functioning.

https://github.com/newrelic/newrelic-cli/blob/master/internal/config/config_integration_test.go

### Go Version
`go version go1.14 darwin/amd64`

### Current behavior
Running the config integration test modifies the user config.yml

### Expected behavior
Tests do not change the user config

### Steps To Reproduce
1. `make test-integration`
2. review ~/.newrelic/config.yml
3. See `logLevel` and `pluginDir` have changed

ctrombley:
Resolves #99 .
 
This PR fixes some internal config logic so that the tests no longer overwrite the actual configuration file.

### DIFF
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -83,7 +83,7 @@ func init() {
 // LoadConfig loads the configuration from disk, substituting defaults
 // if the file does not exist.
 func LoadConfig(configDir string) (*Config, error) {
-	log.Debug("loading config file")
+	log.Debugf("loading config file from %s", configDir)
 
 	if configDir == "" {
 		configDir = DefaultConfigDirectory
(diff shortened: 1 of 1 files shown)


internal/config/config.go
```diff
@@ -402,10 +404,8 @@ func readConfig(configDir string) (*viper.Viper, error) {
 	cfgViper.SetEnvPrefix(DefaultEnvPrefix)
 	cfgViper.SetConfigName(DefaultConfigName)
 	cfgViper.SetConfigType(DefaultConfigType)
-	cfgViper.AddConfigPath(configDir)              // adding provided directory as search path
-	cfgViper.AddConfigPath(DefaultConfigDirectory) // adding home directory as search path
```
jthurman42: should we search these other paths still? If someone passes in a dir that doesn't exist / is empty, and has a default config would it be an error? Maybe an unexpected success?

ctrombley: The problem is that we need a deterministic place to write the file.__

PR #109, opened 2020-03-05
Split sub-commands into files

jthurman42:
Split out each command from `command.go` to `command_subCommandName.go` (tests too)

Ensure all tests are either Unit or Integration


zlesnr:
At first glance my thought is that we're doing too much file splitting.  I think some amount of name-spacing the variables makes sense, so the variable name would indicate the command structure, but in terms of files having the `command.go` has been sort of nice.

jthurman42:
When looking to make some changes to the sub-commands, I found it confusing to have all of the different items in a single large file.

The goal here was to break each sub-command out into it's own self contained file to increase readability and make adding a new sub-command easy.  Since we are using `init()` to register the sub-commands it's literally adding a new file (plus tests) for each new command.

PR #118, opened 2020-03-07
chore(homebrew): add admin build target for homebrew updates

zlesnr:


build/admin.mk
```diff
@@ -0,0 +1,33 @@
...
+	@mkdir -p $(OUTPUT_DIR)/homebrew
+	@pushd $(OUTPUT_DIR)/homebrew \
+		; git checkout master \
+		; git fetch origin master \
+		; git reset --hard origin/master  \
+		; git checkout -b newrelic_$(LATEST_VER) \
+		; popd
+	@pushd $(OUTPUT_DIR)/homebrew; git checkout -b newrelic_$(LATEST_VER);popd
+	@sed -i '' -e 's/v[0-9]*\.[0-9]*\.[0-9]*/$(LATEST_VER)/g' \
+		output/homebrew/Formula/newrelic-cli.rb
```
jthurman42: If you make references to the formula name a VAR, and scrape that out of the environment or a config file, we could port this back to newrelic/developer-toolkit-template-go repo

zlesnr: Oh good idea.

PR #131, opened 2020-03-11
Update newrelic-client-go 0.16.0

jthurman42:
* Update newrelic-client-go to get more metatdata on entities (and fix missing applicationId on apm/application
* Moved AccountID and ApplicationID to the root `apm` command as persistent flags for sub-commands.
* Fix `apm application search` to accept accountId as a filter
* Fix `apm application search` to accept a GUID

### DIFF
--- a/internal/apm/command_application.go
+++ b/internal/apm/command_application.go
@@ -105,12 +105,11 @@ The get command performs a query for an APM application by GUID.
 func init() {
 	Command.AddCommand(cmdApp)
 
-	cmdApp.PersistentFlags().IntVarP(&appID, "applicationId", "a", 0, "search for results matching the given APM application ID")
 	cmdApp.PersistentFlags().StringVarP(&appGUID, "guid", "g", "", "search for results matching the given APM application GUID")
 
 	cmdApp.AddCommand(cmdAppGet)
 
 	cmdApp.AddCommand(cmdAppSearch)
 	cmdAppSearch.Flags().StringVarP(&appName, "name", "n", "", "search for results matching the given APM application name")
-	cmdAppGet.Flags().StringVarP(&appAccountID, "accountId", "", "", "search for results matching the given APM application account ID")
+	cmdAppSearch.Flags().StringVarP(&appAccountID, "accountId", "", "", "search for results matching the given APM application account ID")
 }
--- a/internal/apm/command_application.go
+++ b/internal/apm/command_application.go
@@ -15,10 +15,8 @@ import (
 )
 
 var (
-	appAccountID string
-	appID        int
-	appName      string
-	appGUID      string
+	appName string
+	appGUID string
 )
 
 // Command represents the apm command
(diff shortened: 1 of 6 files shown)


internal/apm/command_application.go
```diff
@@ -105,12 +105,11 @@ The get command performs a query for an APM application by GUID.
 func init() {
 	Command.AddCommand(cmdApp)
 
-	cmdApp.PersistentFlags().IntVarP(&appID, "applicationId", "a", 0, "search for results matching the given APM application ID")
```
ctrombley: No more application ID needed for these commands?

internal/apm/command_deployment.go
```diff
@@ -39,8 +40,13 @@ The list command returns deployments for a New Relic APM application.
 `,
 	Example: "newrelic apm deployment list --applicationId <appID>",
 	Run: func(cmd *cobra.Command, args []string) {
+		if apmAppID == 0 {
```
ctrombley: I wonder if we can just `cmdDeploymentList.MarkFlagRequired("applicationID")`?  Even though it's a persistent flag?

internal/apm/command_deployment.go
```diff
@@ -61,8 +67,13 @@ application.
 `,
 	Example: "newrelic apm deployment create --applicationId <appID> --revision <deploymentRevision>",
 	Run: func(cmd *cobra.Command, args []string) {
+		if apmAppID == 0 {
```
ctrombley: Same as above

internal/apm/command_deployment.go
```diff
@@ -82,8 +93,13 @@ The delete command performs a delete operation for an APM deployment.
 `,
 	Example: "newrelic apm deployment delete --applicationId <appID> --deploymentID <deploymentID>",
 	Run: func(cmd *cobra.Command, args []string) {
+		if apmAppID == 0 {
```
ctrombley: Same as above

internal/apm/command_deployment_test.go
```diff
@@ -20,15 +20,15 @@ func TestApmDeploymentList(t *testing.T) {
 	assert.Equal(t, "list", cmdDeploymentList.Name())
 
 	testcobra.CheckCobraMetadata(t, cmdDeploymentList)
-	testcobra.CheckCobraRequiredFlags(t, cmdDeploymentList, []string{"applicationId"})
+	testcobra.CheckCobraRequiredFlags(t, cmdDeploymentList, []string{})
```
ctrombley: Add these back in if the suggestion above works?

internal/apm/command_deployment.go
jthurman42: You can't since it's a persisted flag, and while there is a MarkPersistentFlagRequired func, reading through the docs / issues say it was never actually implemented to enforce a flag as required 😢

PR #179, opened 2020-04-07
chore(client): Sanitize Region override

### LINKED ISSUE #101: Region selection using NEW_RELIC_REGION is not validated
### Description
If I export the `NEW_RELIC_REGION=test`, I get no error or indication that the US region is being used.

### Steps To Reproduce

```
NEW_RELIC_REGION=test ./bin/darwin/newrelic entity search --name WebPortal
```

Should get an error or warning, instead you get success if the key is valid in the US region (Default)

jthurman42:
Add logic in the overrides function to ensure the regions are valid before continuing.  Region attempt order is ENV, Profile, Default with error messages on every failure.

This should resolve #101

### DIFF
--- a/internal/client/client_integration_test.go
+++ b/internal/client/client_integration_test.go
@@ -1,0 +1,87 @@
+// +build integration
+
+package client
+
+import (
+	"io/ioutil"
+	"os"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+
+	"github.com/newrelic/newrelic-client-go/pkg/region"
+
+	"github.com/newrelic/newrelic-cli/internal/credentials"
+)
+
+var overrideEnvVars = []string{
+	"NEW_RELIC_API_KEY",
+	"NEW_RELIC_REGION",
+}
+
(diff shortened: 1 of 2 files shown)


internal/client/client_integration_test.go
```diff
@@ -0,0 +1,87 @@
...
+
+	// Override just the API Key
+	os.Setenv("NEW_RELIC_API_KEY", "anotherAPIKey")
+	p2 = applyOverrides(&p)
+	assert.NotNil(t, p2)
+	assert.Equal(t, "anotherAPIKey", p2.APIKey)
+	assert.Equal(t, p.Region, p2.Region)
+
+	// Both
+	os.Setenv("NEW_RELIC_REGION", "US")
```
zlesnr, line 61: Perhaps too we should test something that isn't the US, since its the default and could give us false results.

jthurman42: The goal is to check that the env overrides happen, vs testing that Region parsing is working correctly (which happens in `newrelic-client-go`).  That said, I can change one to `eu`

PR #181, opened 2020-04-08
Region fix (again)

jthurman42:
Region is still broken when stored in lower-case (nr1 format) with a confusing error:

`FATAL[] unable to create New Relic client with error: unknown region: us, using default: US`

This adds a custom decoder so when the profiles are loaded, we run the region through `region.Parse` to validate them.

### DIFF
--- a/internal/credentials/profiles.go
+++ b/internal/credentials/profiles.go
@@ -5,8 +5,10 @@ import (
 	"fmt"
 	"io/ioutil"
 	"os"
+	"reflect"
 	"strings"
 
+	"github.com/mitchellh/mapstructure"
 	log "github.com/sirupsen/logrus"
 	"github.com/spf13/viper"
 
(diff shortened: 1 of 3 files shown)


internal/credentials/profiles.go
```diff
@@ -120,7 +122,16 @@ func readCredentials(configDir string) (*viper.Viper, error) {
...
 	cfgMap := map[string]Profile{}
-	err := cfgViper.Unmarshal(&cfgMap)
+
+	// Have to pass in the default hooks to add one...
+	err := cfgViper.Unmarshal(&cfgMap,
+		viper.DecodeHook(
+			mapstructure.ComposeDecodeHookFunc(
+				mapstructure.StringToTimeDurationHookFunc(),
+				mapstructure.StringToSliceHookFunc(","),
+				StringToRegionHookFunc(), // Custom parsing of Region on unmarshal
```
zlesnr, line 132: Oh neat.

PR #188, opened 2020-04-10
Output refactor + formatting options

jthurman42:
Previously, we had one method of output: Pretty Print JSON.  This is fine for folks that are interacting with the CLI, but more options would be better.

This PR:
* Consolidates all output to a new `output` package (creative right?)
* Adds support for YAML output.  Since the data is structured, it was a logical choice though not sure it will get much use
* Adds `--format (json|yaml)` flag that allow runtime changes of the format
* Enables plain / compact output of JSON via `--plain` flag

### DIFF
--- a/internal/output/yaml.go
+++ b/internal/output/yaml.go
@@ -1,0 +1,43 @@
+package output
+
+import (
+	"errors"
+	"fmt"
+
+	"gopkg.in/yaml.v2"
+)
+
+// yaml prints out data as yaml
+func (o *Output) yaml(data interface{}) error {
+	var (
+		formatted []byte
+		err       error
+	)
+
+	// Early quit on no data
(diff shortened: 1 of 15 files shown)


internal/output/yaml.go
```diff
@@ -0,0 +1,43 @@
...
+	if o == nil || o.jsonFormatter == nil {
+		return errors.New("invalid output formatter")
+	}
+
+	// Let's see what they sent us
+	switch d := data.(type) {
+	//case *bytes.Buffer:
+	//	formatted, err = o.jsonFormatter.Format(d.Bytes())
+	//case []byte:
+	//	formatted, err = o.jsonFormatter.Format(d)
```
sanderblue, line 31: Does this need to be uncommented?

Issue #166, opened 2020-04-05
Streamline release process for updating and publishing Homebrew Formula


sanderblue:
**Disclaimer:** Not sure if the GoReleaser part has been attempted yet.

### Objective
For a more efficient build and distribution pipeline, we should automate some of the manual steps we currently have in place. For example, we manually update the New Relic Homebrew [formula file](https://github.com/newrelic-forks/homebrew-core/blob/master/Formula/newrelic-cli.rb) and create a pull request to `homebrew-core` repo from our fork.

#### Utilize GoReleaser if possible
GoReleaser provides the ability to [generate and publish a Homebrew tap](https://goreleaser.com/homebrew/) recipe into a repository that you have access to. In this case, we might be able to integrate this with our fork `newrelic-forks/homebrew-core` repository and have the formula committed directly to that repo from our pipeline. 

##### Examples discovered during research
- CircleCI has a [script](https://github.com/CircleCI-Public/circleci-cli/blob/master/.circleci/brew-deploy.sh) that automatically generates a new PR to the Homebrew repo. Here is the [`brew-deploy`](https://github.com/CircleCI-Public/circleci-cli/blob/master/.circleci/config.yml#L226-L244) section of their config. And here is the relevant [`workflows`](https://github.com/CircleCI-Public/circleci-cli/blob/master/.circleci/config.yml#L262-L269) section of the config.

#### Plan B
If GoReleaser won't work, we can write a script to automate this process and run the script in CircleCI. 

### Acceptance Criteria
- Where possible, manual steps of publishing a new release to Homebrew have been automated
- Any relevant docs should reflect the latest information for installing the CLI via Homebrew

PR #192, opened 2020-04-14
feat(ci): additional automation for homebrew releases

sanderblue:
Resolves: #166

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -48,7 +48,7 @@ jobs:
       - checkout
       - run: git fetch --tags
       - save_cache:
-          key: repo-{{ .Environment.CIRCLE_SHA1 }}
+          key: repo-{{ .Environment.CIRCLE_BRANCH }}-{{ .Environment.CIRCLE_SHA1 }}
           paths:
             - ~/project
 
@@ -57,7 +57,7 @@ jobs:
       - image: circleci/golang:1.14
     steps:
       - restore_cache:
-          key: repo-{{ .Environment.CIRCLE_SHA1 }}
+          key: repo-{{ .Environment.CIRCLE_BRANCH }}-{{ .Environment.CIRCLE_SHA1 }}
       - run:
           name: Lint
           command: make lint
@@ -67,7 +67,7 @@ jobs:
       - image: circleci/golang:1.14
     steps:
       - restore_cache:
-          key: repo-{{ .Environment.CIRCLE_SHA1 }}
+          key: repo-{{ .Environment.CIRCLE_BRANCH }}-{{ .Environment.CIRCLE_SHA1 }}
       - run:
           name: Unit Tests
           command: make test-unit
@@ -80,7 +80,7 @@ jobs:
       - image: circleci/golang:1.14
     steps:
       - restore_cache:
-          key: repo-{{ .Environment.CIRCLE_SHA1 }}
+          key: repo-{{ .Environment.CIRCLE_BRANCH }}-{{ .Environment.CIRCLE_SHA1 }}
       - snyk/scan:
           severity-threshold: high
           fail-on-issues: false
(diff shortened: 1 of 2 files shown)


scripts/newrelic-cli.rb.tmpl
```diff
@@ -0,0 +1,34 @@
+class NewrelicCli < Formula
```
jthurman42, line 1: A better place for this file would be `build/package/homebrew/`

.circleci/config.yml
```diff
@@ -101,5 +101,14 @@ jobs:
       - checkout
       - setup_remote_docker:
           docker_layer_caching: true
-      - run: make release-publish
-
+      - run:
+          name: 'Set custom environment variables'
+          command: |
+            echo 'export GIT_TAG=$(git describe --tags | tr -d "v")' >> $BASH_ENV
+            source $BASH_ENV
```
jthurman42, line 108: Don't we get the latest git tag from Circle already? `CIRCLE_TAG` We could short-cut this to aliasing that env var.

https://circleci.com/docs/2.0/env-vars/#built-in-environment-variables

.circleci/config.yml
```diff
@@ -101,5 +101,14 @@ jobs:
...
+      - run:
+          name: 'Generate release and publish'
+          command: make release-publish
+      - run:
+          name: 'Generate homebrew-core pull request'
+          command: ./scripts/brew-formula.sh
```
jthurman42, line 114: Goreleaser also supports post-build hooks so this could be part of the process above.  It should be part / called via `make release-publish` if you don't want to use the post-build hooks.  https://goreleaser.com/build/

scripts/newrelic-cli.rb.tmpl
sanderblue: Won't be using this anymore, but good to know where to put those sorts of things 😎

.circleci/config.yml
sanderblue: I was having issues with passing the built-in environment variables to the bash script. Then I found some article/post that CircleCI won't pass the built-in variables to scripts for security reasons, so this is their prescribed workaround. Now that I have things stable, I can give the built-in env vars another try to see if I just missed something earlier on.

sanderblue: Closing the loop. Here's the fine print regarding what I mentioned for the built-in env vars, which is part of the docs section linked above.

> Note: You cannot use a built-in environment variable to define another environment variable. Instead, you must use a `run` step to export the new environment variables using `BASH_ENV`.

Seems a little odd, but at least what we've currently got works.

.circleci/config.yml
sanderblue: I'll take a look at this today.

sanderblue: 🗒️: To close the loop, it turns out GoReleaser's `post` hook within `build` doesn't contain the archived assets so we can't run the script via the `build.hooks.post` hook. And although there's a global top-level `before.hooks`, there is no global top-level `after.hooks`. It would be ideal to have the global `after` hook as well as a `post` hook within the `archives` configuration.  ¯\\_(ツ)_/¯

scripts/brew-formula.sh
```diff
@@ -0,0 +1,72 @@
...
+
+# Set variables for lines to replace in the formula (lines 4 and 5)
+formula_url='  url "https:\/\/github.com\/newrelic\/newrelic-cli\/archive\/v'${GIT_TAG}'.tar.gz"'
+formula_sha256='  sha256 "'${SHA256}'"'
+
+# Make temporary copy of existing formula file
+cp $homebrew_formula_file $tmp_formula_file
+
+# Replace lines 4 and 5 in the formula file, using the .tmp file as a template
+sed -e '4s/.*/'"${formula_url}"'/' -e '5s/.*/'"${formula_sha256}"'/' $tmp_formula_file > $homebrew_formula_file
```
zlesnr, line 57: Nice.

zlesnr: If you wanted to avoid making a copy of the file, GNU `sed -i -e...` will make the changes in place.  BSD sed on mac requires an extension, but you can get the same behavior with `sed -i '' -e...`.  What you gave here looks good to me though.

PR #193, opened 2020-04-14
feat(snapcraft): include goreleaser config for snaps

zlesnr:


build/snapcraft.mk
```diff
@@ -0,0 +1,16 @@
+#
+# Makefile fragment for Snapcraft actions
+#
+SNAPCRAFT         ?= snapcraft
+
+snapcraft-login:
```
jthurman42, line 6: Does this need to be called as a dep of `make release-publish`?

zlesnr: Oh yes, thank you.

PR #213, opened 2020-04-22
refactor(workload): query with GUID per upstream API change

sanderblue:
Depends on: [newrelic-client-go PR #261](https://github.com/newrelic/newrelic-client-go/pull/261)


jthurman42:
Resolved in #220

PR #217, opened 2020-04-23
feat(packaging): include rpm and deb builds in goreleaser

zlesnr:


.goreleaser.yml
```diff
@@ -116,3 +116,29 @@ scoop:
...
+    homepage: https://github.com/newrelic/newrelic-cli
+    maintainer: opensource@newrelic.com
+    description: |
+      The New Relic CLI is an officially supported command line interface for New
+      Relic, released as part of the Developer Toolkit.
+    license: Apache 2.0
+    formats:
+      - deb
+      - rpm
+    bindir: /usr/bin
```
jthurman42, line 142: Do we need to override this from `/usr/local/bin`? Or is this for consistency with the snapcraft / others?

zlesnr: I think snapcraft is all `/usr/local/bin`, so I can make it match here.

PR #267, opened 2020-05-13
fix(build): ignore Scoop commits during commit linting

ctrombley:


build/lint.mk
```diff
@@ -8,7 +8,7 @@ GOFMT        ?= gofmt
 GOIMPORTS    ?= goimports
 
 COMMIT_LINT_CMD   ?= go-gitlint
-COMMIT_LINT_REGEX ?= "(chore|docs|feat|fix|refactor|tests?)(\([^\)]+\))?: .*"
+COMMIT_LINT_REGEX ?= "(Scoop update|(chore|docs|feat|fix|refactor|tests?)(\([^\)]+\))?:) .*"
```
sanderblue, line 11: Did we adjust the commit message Scoop makes? @zlesnr's PR was merged and released I believe, so we should be able to use a commit message template for Scoop in our GoReleaser config.

zlesnr: The change to the goreleaser config has been made, but some issue remains.

ctrombley: We did that and somehow we are still getting the old style commits which are now blocking releases.  I know this isn't a "fix-forward" type change that we usually like to do but I wanted to unblock releases while we figure out what's going on.

sanderblue: Totally agreed with this approach for this kind of scenario :)

PR #295, opened 2020-05-28
[feat] Add pre-release config option

### LINKED ISSUE #274: Pre-GA Feature flag
### Feature Description
As we add support for features that are not yet generally available (though publicly announced) we should have a way to hide them to not clutter up the CLI, but expose them in a way that those who opt-in will see them.

### Describe Alternatives
We could also add a simple text snippet to the commands, and let them be exposed.  We should add some contextual help for them anyway on how to engage with New Relic to evaluate a new feature.

### Additional context
Edge tracing is a great example of a use-case for this flow.

jthurman42:
We want to ship features that have been announced, but not GA in a way that is not confusing to our users.  To do so, we will add a config option to display the non-GA features within the CLI if a user chooses to do so.  Otherwise hide them by default.

This resolves #274

### DIFF
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -26,9 +26,6 @@ const (
 	// DefaultLogLevel is the default log level
 	DefaultLogLevel = "INFO"
 
-	// DefaultSendUsageData is the default value for sendUsageData
-	DefaultSendUsageData = "NOT_ASKED"
-
 	globalScopeIdentifier = "*"
 )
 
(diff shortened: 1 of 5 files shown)


internal/config/config_integration_test.go
```diff
@@ -47,11 +47,11 @@ func TestConfigSetSendUsageData(t *testing.T) {
 	assert.NoError(t, err)
 	assert.Equal(t, c.configDir, f)
 
-	// Set the valid log levels
-	for _, l := range []string{
-		"NOT_ASKED",
-		"DISALLOW",
-		"ALLOW",
+	// Set the valid send-usage-date values
```
sanderblue, line 50: lil typo?
```diff
- send-usage-date
+ send-usage-data
```

ctrombley, line 50: nit: s/date/data

internal/config/config.go
```diff
@@ -42,9 +39,10 @@ var (
 
 // Config contains the main CLI configuration
 type Config struct {
-	LogLevel      string `mapstructure:"logLevel"`      // LogLevel for verbose output
-	PluginDir     string `mapstructure:"pluginDir"`     // PluginDir is the directory where plugins will be installed
-	SendUsageData string `mapstructure:"sendUsageData"` // SendUsageData enables sending usage statistics to New Relic
+	LogLevel           string  `mapstructure:"logLevel"`           // LogLevel for verbose output
+	PluginDir          string  `mapstructure:"pluginDir"`          // PluginDir is the directory where plugins will be installed
+	SendUsageData      Ternary `mapstructure:"sendUsageData"`      // SendUsageData enables sending usage statistics to New Relic
+	PreReleaseFeatures Ternary `mapstructure:"PreReleaseFeatures"` // PreReleaseFeatures enables display on features within the CLI that are announced but not generally available to customers
```
ctrombley, line 45: nit: camel case for consistency?

PR #306, opened 2020-06-09
Query NRQL via the CLI

jthurman42:
Support for NRQL Queries from the CLI:

* Query NRDB: `newrelic nrql query --query 'FROM Transaction SELECT count(*)' --accountId 123456789`
* Show query history: `newrelic nrql history`

This also consolidates some of the Text / terminal based output into the `output` package, and removed some deps in the process.  Note that the exact text representation for config/credentials is NOT the same visually.


internal/nrql/command_query.go
```diff
@@ -0,0 +1,67 @@
...
+			return errors.New("command expects only 1 argument")
+		}
+
+		return nil
+	},
+	Run: func(cmd *cobra.Command, args []string) {
+		client.WithClient(func(nrClient *newrelic.NewRelic) {
+			queryVars := map[string]interface{}{
+				"accountId": accountID,
+				"query":     args[0],
```
jthurman42, line 50: The NerdGraph resource is the only other place we do this (use args vs flags), so going to change that.

jthurman42:
Blocked on newrelic/newrelic-client-go#357

internal/config/config.go
```diff
@@ -151,13 +153,14 @@ func (c *Config) Delete(key string) error {
 		return err
 	}
 
-	renderer.Delete(key)
+	output.Printf("%s %s removed successfully\n", text.FgGreen.Sprint("✔"), text.Bold.Sprint(key))
```
sanderblue, line 156: This is a nice UX touch 👍

internal/output/print.go
```diff
@@ -9,6 +15,8 @@ func Print(data interface{}) (err error) {
 	switch globalOutput.format {
 	case FormatJSON:
 		err = globalOutput.json(data)
+	case FormatText:
+		err = globalOutput.table(data)
```
sanderblue, line 19: `FormatText ` == `.table(data)`? This is probably a semantic thing, but "text" and "table" feel like different output types to me.

jthurman42: That's totally a bug, should be `.text`, standby. (Was FormatTable previously)

PR #325, opened 2020-07-02
fix(config): ensure compatible case with nr1

### LINKED ISSUE #120: profile uses inconsistent casing for API: adminAPIKey, apiKey
### Objective
Clean up the variable casing here.  Maintain compatibility with `nr1`.

SowmyaJujjuri:
Resolved #120

### DIFF
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -27,7 +27,7 @@ const (
 	DefaultEnvPrefix = "newrelic"
 
 	// DefaultLogLevel is the default log level
-	DefaultLogLevel = "INFO"
+	DefaultLogLevel = "Info"
 
 	globalScopeIdentifier = "*"
 )
@@ -166,7 +166,7 @@ func (c *Config) Get(key string) {
 // Set sets a config value.
 func (c *Config) Set(key string, value interface{}) error {
 	if !stringInStrings(key, validConfigKeys()) {
-		return fmt.Errorf("\"%s\" is not a valid key; Please use one of: %s", key, validConfigKeys())
+		return fmt.Errorf("\"%s\" is not a valid key; use one of: %s", key, validConfigKeys())
 	}
 
 	err := c.set(key, value)
@@ -438,7 +438,7 @@ func validConfigKeys() []string {
 
 func stringInStrings(s string, ss []string) bool {
 	for _, v := range ss {
-		if strings.EqualFold(v, s) {
+		if v == s {
 			return true
 		}
 	}
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -335,7 +335,6 @@ func (c *Config) setDefaults() error {
 
 func (c *Config) validate() error {
 	err := c.visitAllConfigFields(func(v *Value) error {
-		// switch k := strings.ToLower(v.Name); k {
 		switch k := strings.ToLower(v.Name); k {
 		case "loglevel":
 			validValues := []string{"Info", "Debug", "Trace", "Warn", "Error"}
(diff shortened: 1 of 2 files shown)


internal/config/config.go
```diff
@@ -166,7 +166,7 @@ func (c *Config) Get(key string) {
 // Set sets a config value.
 func (c *Config) Set(key string, value interface{}) error {
 	if !stringInStrings(key, validConfigKeys()) {
-		return fmt.Errorf("\"%s\" is not a valid key; Please use one of: %s", key, validConfigKeys())
+		return fmt.Errorf("\"%s\" is not a valid key; use one of: %s", key, validConfigKeys())
```
zlesnr, line 169: Do we still want to say please here?

internal/config/config.go
```diff
@@ -438,7 +438,7 @@ func validConfigKeys() []string {
 
 func stringInStrings(s string, ss []string) bool {
 	for _, v := range ss {
-		if strings.EqualFold(v, s) {
+		if v == s {
```
ctrombley, line 441: This will enforce case sensitivity when validating keys, which is an effective solution for this issue.  Nice job!

However, `stringInStrings` is also used to [validate the config *values*](https://github.com/newrelic/newrelic-cli/blob/master/internal/config/config.go#L339-L349), so case sensitivity would be enforced when checking a log level the user has chosen.  This is less user friendly - we want them to be able to use "INFO", "info", or "Info" for instance.   

How about introducing a boolean arg similar to [this code](https://github.com/hashicorp/terraform-plugin-sdk/blob/master/helper/validation/strings.go#L132) that would allow turning case sensitivity on or off depending on the use case?  We could use the case sensitive version to validate the keys, and the case insensitive version to validate the values.

We will also need to update the case of the keys in the [`validate` method](https://github.com/newrelic/newrelic-cli/blob/master/internal/config/config.go#L339-L349) after enforcing case sensitivity, as those case statements will stop evaluating to true.

internal/config/config_integration_test.go
```diff
@@ -22,11 +22,11 @@ func TestConfigSetLogLevel(t *testing.T) {
 
 	// Set the valid log levels
 	for _, l := range []string{
-		"ERROR",
-		"WARN",
-		"INFO",
-		"DEBUG",
-		"TRACE",
+		"Error",
```
ctrombley, line 25: See the related comment.  We should still allow the user to use case-insensitive values.

internal/config/config.go
```diff
@@ -27,7 +27,7 @@ const (
 	DefaultEnvPrefix = "newrelic"
 
 	// DefaultLogLevel is the default log level
-	DefaultLogLevel = "INFO"
+	DefaultLogLevel = "Info"
```
ctrombley, line 30: See the related comment.  We should still allow the user to use case-insensitive values.

internal/config/config_integration_test.go
```diff
@@ -35,6 +35,12 @@ func TestConfigSetLogLevel(t *testing.T) {
 
 	err = c.Set("logLevel", "INVALID_VALUE")
 	assert.Error(t, err)
+
+	err = c.Set("loglevel", "Info")
+	assert.Error(t, err)
+
+	err = c.Set("Loglevel", "Debug")
+	assert.Error(t, err)
```
zlesnr, line 43: What do you think about also including a couple tests for a `c.Get()` just to ensure that we're getting back what we expect also.  I think that would probably effectively test the round trip here.

internal/config/config.go
```diff
@@ -335,10 +335,11 @@ func (c *Config) setDefaults() error {
 
 func (c *Config) validate() error {
 	err := c.visitAllConfigFields(func(v *Value) error {
+		// switch k := strings.ToLower(v.Name); k {
```
ctrombley, line 338: Can this be removed?

PR #345, opened 2020-07-24
feat(reporting): add junit reporting

ctrombley:
Add JUnit reporting to NRDB via the `newrelic reporting junit` command.

This PR also adds JUnit test output to the build pipeline, so it can be used for testing of this functionality.  

### Testing
```
./bin/darwin/newrelic reporting junit --accountId 2520528 --path ./coverage/integration.xml
```

### DIFF
--- a/internal/reporting/command_junit.go
+++ b/internal/reporting/command_junit.go
@@ -1,0 +1,109 @@
+package reporting
+
+import (
+	"io/ioutil"
+
+	"github.com/google/uuid"
+	"github.com/joshdk/go-junit"
+	log "github.com/sirupsen/logrus"
+	"github.com/spf13/cobra"
+
+	"github.com/newrelic/newrelic-cli/internal/client"
+	"github.com/newrelic/newrelic-cli/internal/credentials"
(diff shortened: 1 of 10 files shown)


internal/reporting/command_junit.go
```diff
@@ -0,0 +1,109 @@
...
+	}
+
+	return e
+}
+
+func init() {
+	Command.AddCommand(cmdJUnit)
+	cmdJUnit.Flags().IntVarP(&accountID, "accountId", "a", 0, "the New Relic account ID to send test run results to")
+	cmdJUnit.Flags().StringVarP(&path, "path", "p", "", "the path to a JUnit-formatted test results file")
+	cmdJUnit.Flags().BoolVarP(&convertOnly, "convertOnly", "c", false, "convert to custom event format without posting events")
```
jthurman42, line 106: I think `--dry-run` would be a better name here (without the short option)

sanderblue: +1

ctrombley: 👍

PR #355, opened 2020-08-10
chore(deps): upgrade newrelic-client-go

zlesnr:


internal/client/client_integration_test.go
```diff
@@ -62,26 +62,26 @@ func TestApplyOverrides(t *testing.T) {
...
-	assert.Equal(t, region.EU, p2.Region)
+	assert.Equal(t, region.EU.String(), p2.Region)
 
 	// Invalid REGION (Should be region.Default)
-	os.Setenv("NEW_RELIC_REGION", "test")
-	p2 = applyOverrides(&p)
-	assert.NotNil(t, p2)
-	assert.Equal(t, p.APIKey, p2.APIKey)
-	assert.Equal(t, region.Default, p2.Region)
+	// os.Setenv("NEW_RELIC_REGION", "test")
```
zlesnr, line 82: Now that region is just a string, we're not doing any validation calls, so here I just commented it out.  Is there something else we'd prefer to do here, or should I remove this test?

ctrombley: We can just remove it.  It's tested in the client.

zlesnr: Deal, removed.

PR #364, opened 2020-08-26
feat: Add Config Value Obfuscation utility for Agents

jifeingo:
* Adds new `agent` area supporting agent specific functions
    * Adds new `config` area supporting configuration helper functions
    * Adds new command line support for obfuscating configuration value using a key
example invocation: `newrelic agent config obfuscate --value XYZ --key 123`
    * Adds new utility helper functions
    * Unit Tests of command hierarchy and obfuscation logic.


internal/agent/command.go
```diff
@@ -0,0 +1,12 @@
+package agent
+
+import (
+	"github.com/spf13/cobra"
+)
+
+// Command represents the apm command
```
ctrombley, line 7: nit: s/apm/agent/

internal/agent/command_config.go
```diff
@@ -0,0 +1,48 @@
...
+
+var cmdConfig = &cobra.Command{
+	Use:   "config",
+	Short: "Configuration utilities/helpers for New Relic agents",
+}
+
+var cmdConfigObfuscate = &cobra.Command{
+	Use:   "obfuscate",
+	Short: "Obfuscate a configuration value using a key",
+	Long: `Obfuscate a configuration value using a key.  Obfuscated value
```
ctrombley, line 23: Suggestion: s/is placed/should be used/

internal/agent/command_config.go
```diff
@@ -0,0 +1,48 @@
...
+	},
+}
+
+func init() {
+
+	Command.AddCommand(cmdConfig)
+
+	cmdConfig.AddCommand(cmdConfigObfuscate)
+
+	cmdConfigObfuscate.Flags().StringVarP(&encodeKey, "key", "", "", "the key to use when obfuscating the clear-text value")
```
ctrombley, line 43: The 3rd argument to `StringVarP` is a single letter to provide a short option for the argument (e.g. `--key` vs `-k`).  "k" and "v" might be good values for these two arguments.  If you're not interested in a short form options, you can just use `StringVar`.

internal/agent/utils.go
```diff
@@ -0,0 +1,35 @@
+package agent
+
+import "encoding/base64"
+
+// ObfuscationResult represents the result of an obfuscation request
+type ObfuscationResult struct {
+	ObfuscatedValue string
```
ctrombley, line 7: We try and stick to camel case for JSON output values.  You can achieve that here by adding a struct tag:

```go
type ObfuscationResult struct {
	ObfuscatedValue string `json:"obfuscatedValue"`
}
```

Issue #45, opened 2020-02-21
Accept input via command chaining


jthurman42:
### Feature Description
Add the ability to chain newrelic commands together, example would be an entity search filtered to GUIDs, then piped back to newrelic to add tags to all of those entities.

```
$ newrelic entity search --name "mystuff" --only-attributes entityId | \
  newrelic entity tags add --tag owner:me
```

nicolasjhampton:
@jthurman42 I intend to start tackling this issue, using the entity command as an MVP, but it would help if you could answer a question. This example code includes a secret request to add a feature for bulk edits. Currently in the cli, there's no way to specify multiple entities to act on at once. Should I aim here to add that functionality?

jthurman42:
@nicolasjhampton Good point, there is an assumption that the CLI would handle an array of input items and loop thought them based on the example.  I believe my thought here was that the data passed through the pipe would be in JSON, so easy enough to detect an array to iterate on.  I'd prefer to see that at least though of, otherwise we're pushing the work of splitting / running multiple items out to the user.

nicolasjhampton:
@jthurman42 I've gotten all the way to getting bulk information from stdin to the actual newrelic client query, but there's literally not a method on the newrelic-client-go package that takes several guids and returns the tags for all of them. 

Like we agreed before, this is actually two legs of work. I think I'm complete on getting bulk data from stdin and making it accessible to each command to use. If I can get your signoff for this to be reviewed and merged separately, I'll do an inventory of what needs to change in newrelic-client-go to allow useful bulk actions and make a PR there to make further work here possible.

Julien4218:
cc @kidk

PR #368, opened 2020-08-31
feat(release): Issue #45 add support for command chaining

nicolasjhampton:
# What

This PR adds a pipe module to read stdin for piping input to commands, allowing each command
be piped to the next. I've also implemented this pipe logic for one command, the `entity tags get`
command. You can try this functionality by running:

`newrelic entity search --name [app name] | newrelic entity tags get`

# Next work needed

Currently the above action only returns tags for the first entity found. Work has to be done in 
newrelic-client-go to add bulk queries to the API. Specifically for this tag query, something like this query needs 
to be added (this is incorrect, working on it in a different repo):

```
query($guids:[EntityGuid]!) {
  actor {
    entities(guids: $guids) {
        tags {
          values
          key
        }
      }
    }
  }
}
```

For the entity tag command alone, nrClient.Entities.ListTags, nrClient.Entities.DeleteTags, nrClient.Entities.DeleteTags, nrClient.Entities.DeleteTagValues, nrClient.Entities.AddTags, and nrClient.Entities.ReplaceTags all need bulk versions written that can take multiple entity guids.

# Documentation

Doc comments included for all of the Public and some of the private API

https://github.com/newrelic/newrelic-cli/issues/45

### DIFF
--- a/internal/entities/command_tags.go
+++ b/internal/entities/command_tags.go
@@ -9,6 +9,7 @@ import (
 
 	"github.com/newrelic/newrelic-cli/internal/client"
 	"github.com/newrelic/newrelic-cli/internal/output"
+	"github.com/newrelic/newrelic-cli/internal/pipe"
 	"github.com/newrelic/newrelic-cli/internal/utils"
 	"github.com/newrelic/newrelic-client-go/newrelic"
 	"github.com/newrelic/newrelic-client-go/pkg/entities"
@@ -40,10 +41,16 @@ The get command returns JSON output of the tags for the requested entity.
 	Example: "newrelic entity tags get --guid <entityGUID>",
 	Run: func(cmd *cobra.Command, args []string) {
 		client.WithClient(func(nrClient *newrelic.NewRelic) {
-			tags, err := nrClient.Entities.ListTags(entityGUID)
-			utils.LogIfFatal(err)
-
-			utils.LogIfError(output.Print(tags))
+			// Temporary until bulk actions can be build into newrelic-client-go
+			if value, ok := pipe.Get("guid"); ok {
+				tags, err := nrClient.Entities.ListTags(value[0])
+				utils.LogIfFatal(err)
+				utils.LogIfError(output.Print(tags))
+			} else {
+				tags, err := nrClient.Entities.ListTags(entityGUID)
+				utils.LogIfFatal(err)
+				utils.LogIfError(output.Print(tags))
+			}
 		})
 	},
 }
@@ -200,8 +207,13 @@ func init() {
 	Command.AddCommand(cmdTags)
 
 	cmdTags.AddCommand(cmdTagsGet)
-	cmdTagsGet.Flags().StringVarP(&entityGUID, "guid", "g", "", "the entity GUID to retrieve tags for")
-	utils.LogIfError(cmdTagsGet.MarkFlagRequired("guid"))
+
+	pipe.GetInput([]string{"guid"})
+
+	if !pipe.Exists("guid") {
+		cmdTagsGet.Flags().StringVarP(&entityGUID, "guid", "g", "", "the entity GUID to retrieve tags for")
+		utils.LogIfError(cmdTagsGet.MarkFlagRequired("guid"))
+	}
 
 	cmdTags.AddCommand(cmdTagsDelete)
 	cmdTagsDelete.Flags().StringVarP(&entityGUID, "guid", "g", "", "the entity GUID to delete tags on")
--- a/internal/pipe/pipe.go
+++ b/internal/pipe/pipe.go
@@ -1,0 +1,162 @@
+// Package pipe provides a simple API to read and retrieve values
+// from stdin to use in Cobra commands. Public API consists of
+// GetInput, which reads stdin, Exists, which checks for value
+// existence, and Get for retrieving existing values.
+package pipe
+
+import (
+	"bufio"
+	"errors"
+	"fmt"
+	"io"
+	"os"
+	"strings"
+
+	"github.com/tidwall/gjson"
+
+	"github.com/newrelic/newrelic-cli/internal/utils"
+)
+
+// Created Interface and struct to surround io.Reader for easy mocking
+type pipeReader interface {
+	ReadPipe() (string, error)
+}
+
+type stdinPipeReader struct {
+	input io.Reader
+}
+
+func (spr stdinPipeReader) ReadPipe() (string, error) {
+	text := ""
+	var err error = nil
+	scanner := bufio.NewScanner(spr.input)
+
+	for scanner.Scan() {
+		text = fmt.Sprintf("%s%s ", text, strings.TrimSpace(scanner.Text()))
+	}
+
+	if scanErr := scanner.Err(); scanErr != nil {
+		err = scanErr
+	}
+
+	return strings.TrimSpace(text), err
+}
+
+func jsonToFilteredMap(r string, selectors []string) ([]map[string]string, error) {
+	text := r
+
+	if !gjson.Valid(text) {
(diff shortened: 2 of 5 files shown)


zlesnr:
Oh interesting, thanks for putting this up @nicolasjhampton.  One idea that @sanderblue and I talked about was having a "mother" parent object that would nest each of the kinds of objects we expect to want to take over STDIN.  In this way, the shape of the received object could change, and each command would know what it would need to expect.  I didn't play with this idea to know if it would work though.

nicolasjhampton:
> Oh interesting, thanks for putting this up @nicolasjhampton. One idea that @sanderblue and I talked about was having a "mother" parent object that would nest each of the kinds of objects we expect to want to take over STDIN. In this way, the shape of the received object could change, and each command would know what it would need to expect. I didn't play with this idea to know if it would work though.

That sounds like it's necessary for encoding/json and most parsers in go. I did the same thing, just without the object list. I got around needing to know the shape by finding a parser that didn't require a struct that knows the shape of the input. That way, as long as it's valid json in the stdin with the attribute you need at the first level of the object, the command can read only the attributes it needs, never having to know the full shape of the data coming in. https://github.com/tidwall/gjson

nicolasjhampton:
If this is a good general direction and there's still buy-in for this work, I'd like to clean up some code, make another helper method or two, add more error handling and tests, and finish out the entity commands as a proof of concept.

internal/pipe/pipe.go
```diff
@@ -0,0 +1,131 @@
...
+	return func(acceptedPipeInput []string) {
+		pipeInput = getPipeInputInnerFunc(pipe, predicate(), acceptedPipeInput)
+	}
+}
+
+var GetInput = getPipeInputFactory(stdinPipeReader{input: os.Stdin}, pipeInputExists)
+
+var pipeInput map[string][]string = map[string][]string{}
+
+func Get(inputKey string) ([]string, bool) {
```
zlesnr, line 123: This method looks to be called once unless I overlooked another place.  What do you think about folding this logic into the `Exists()` method?

internal/pipe/pipe.go
```diff
@@ -0,0 +1,131 @@
...
+
+func getPipeInputFactory(pipe pipeReader, predicate func() bool) func([]string) {
+	return func(acceptedPipeInput []string) {
+		pipeInput = getPipeInputInnerFunc(pipe, predicate(), acceptedPipeInput)
+	}
+}
+
+var GetInput = getPipeInputFactory(stdinPipeReader{input: os.Stdin}, pipeInputExists)
+
+var pipeInput map[string][]string = map[string][]string{}
```
zlesnr, line 121: The logic that populates this variable is a little hard to read.  If multiple `getPipeInputFactory` functions are created, the calls to those methods would potentially clobber this variable and then anything relying on it might go a little pear shaped.  Even in the case of `GetInput()` being called with different arguments I think would change the content of this variable.

zlesnr (changes requested):
This is really neat.  I checked out the code to play with it a bit, and I like what you're doing here.  Some of the methods are hard to read, or to know what they do in the schema of things, and so at least I'd think they need some documentation about their purpose.  I left a few comments, so let me know what you think.

internal/entities/command_tags.go
```diff
@@ -200,8 +207,13 @@ func init() {
 	Command.AddCommand(cmdTags)
 
 	cmdTags.AddCommand(cmdTagsGet)
-	cmdTagsGet.Flags().StringVarP(&entityGUID, "guid", "g", "", "the entity GUID to retrieve tags for")
-	utils.LogIfError(cmdTagsGet.MarkFlagRequired("guid"))
+
+	pipe.GetInput([]string{"guid"})
+
+	if !pipe.Exists("guid") {
```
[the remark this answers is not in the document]
nicolasjhampton: The `GetInput` static method on the Pipe module reads the stdin and stores the values specified by the string array it's given.

internal/entities/command_tags.go
```diff
@@ -40,10 +41,16 @@ The get command returns JSON output of the tags for the requested entity.
 	Example: "newrelic entity tags get --guid <entityGUID>",
 	Run: func(cmd *cobra.Command, args []string) {
 		client.WithClient(func(nrClient *newrelic.NewRelic) {
-			tags, err := nrClient.Entities.ListTags(entityGUID)
-			utils.LogIfFatal(err)
-
-			utils.LogIfError(output.Print(tags))
+			// Temporary until bulk actions can be build into newrelic-client-go
```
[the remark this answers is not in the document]
nicolasjhampton: Most of the API in the newrelic-client-go library aren't designed to deal with multiple entity guids. In this tag command alone, bulk entity commands for nrClient.Entities.ListTags, nrClient.Entities.DeleteTags, nrClient.Entities.DeleteTagValues, nrClient.Entities.AddTags, and nrClient.Entities.ReplaceTags need to be created.

internal/pipe/pipe_test.go
```diff
@@ -0,0 +1,356 @@
...
+						"retail": 75
+					}
+				}
+			]`,
+			"expected": `[ { "id": 1, "name": "Foo", "price": 123, "tags": [ "Bar", "Eek" ], "stock": { "warehouse": 300, "retail": 20 } }, { "id": 2, "name": "Bar", "price": 456, "tags": [ "Oop", "Aah" ], "stock": { "warehouse": 450, "retail": 50 } }, { "id": 3, "name": "Baz", "price": 789, "tags": [ "Syn", "Ack" ], "stock": { "warehouse": 100, "retail": 75 } } ]`,
+		},
+	}
+
+	for _, c := range cases {
+		input := c["input"]
```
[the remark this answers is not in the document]
nicolasjhampton: I like that, then the test cases can be named. I'll work on that.

internal/pipe/pipe.go
nicolasjhampton: The Get method would be called everytime a command needs to get a value. It's called once because I've only used this module on one command. It's just good, clean, self-explanatory API design.

internal/pipe/pipe.go
nicolasjhampton: `getPipeInputFactory` is a lowercase function, meaning it's private to this module and can't be called outside of it. The only time it's ran is when this module is loaded. I do see how GetInput could be ran twice to alter pipeInput, however. I'll add code to prevent pipeInput from being overridden if GetInput is ran twice.

internal/entities/command_tags.go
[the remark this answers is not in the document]
zlesnr: A yes, methods that will take a list of guids to operate on.

internal/pipe/pipe_test.go
```diff
@@ -0,0 +1,463 @@
...
+						"retail": 75
+					}
+				}
+			]`,
+			Expected: `[ { "id": 1, "name": "Foo", "price": 123, "tags": [ "Bar", "Eek" ], "stock": { "warehouse": 300, "retail": 20 } }, { "id": 2, "name": "Bar", "price": 456, "tags": [ "Oop", "Aah" ], "stock": { "warehouse": 450, "retail": 50 } }, { "id": 3, "name": "Baz", "price": 789, "tags": [ "Syn", "Ack" ], "stock": { "warehouse": 100, "retail": 75 } } ]`,
+		},
+	}
+
+	for _, c := range cases {
+		mockStdin := strings.NewReader(c.Input)
```
zlesnr, line 70: This is neat.

zlesnr (approved):
This looks great.  Thanks for adding some method docs too!  Let me know when its no longer WIP and I'll run the tests locally before merge.

zlesnr:
The tests passed in https://github.com/newrelic/newrelic-cli/pull/380.  Merging.

Issue #333, opened 2020-07-14
Panic: NRQL query with text output


jthurman42:
### Description
Table formatting can cause a panic on NRQL results.

### Go Version
CLI version 0.10.0 (Installed via Homebrew)

go version go1.14.4 darwin/amd64

### Current behavior
CLI panics

### Expected behavior
Results would be printed in a table

### Steps To Reproduce

```
$ newrelic nrql query \
    --accountId $NEW_RELIC_ACCOUNT_ID  \
    --query 'FROM NrAuditEvent SELECT count(*) FACET targetType SINCE 1 week ago' \
    --format text


panic: reflect: NumField of non-struct type nrdb.NrdbResult

goroutine 1 [running]:
reflect.(*rtype).NumField(0x1574dc0, 0x16f0ce0)
	/usr/local/Cellar/go/1.14.4/libexec/src/reflect/type.go:973 +0xb7
github.com/newrelic/newrelic-cli/internal/output.(*Output).renderAsTable(0xc0000a6300, 0x152f740, 0xc0001f1140, 0x0, 0x8)
	/private/tmp/newrelic-cli-20200710-99399-1jjc2iq/newrelic-cli-0.10.0/internal/output/text.go:54 +0x16d
github.com/newrelic/newrelic-cli/internal/output.(*Output).text(0xc0000a6300, 0x152f740, 0xc0001f1140, 0x1551f20, 0xc000169301)
	/private/tmp/newrelic-cli-20200710-99399-1jjc2iq/newrelic-cli-0.10.0/internal/output/text.go:28 +0xc7
github.com/newrelic/newrelic-cli/internal/output.Print(0x152f740, 0xc0001f1140, 0x4, 0xc0001f1140)
	/private/tmp/newrelic-cli-20200710-99399-1jjc2iq/newrelic-cli-0.10.0/internal/output/print.go:19 +0xdc
github.com/newrelic/newrelic-cli/internal/nrql.glob..func1.1(0xc000192000)
	/private/tmp/newrelic-cli-20200710-99399-1jjc2iq/newrelic-cli-0.10.0/internal/nrql/command_query.go:38 +0xff
github.com/newrelic/newrelic-cli/internal/client.WithClientFrom.func1.1(0xc00017f110)
	/private/tmp/newrelic-cli-20200710-99399-1jjc2iq/newrelic-cli-0.10.0/internal/client/helpers.go:25 +0xb1
[...]

PR #379, opened 2020-09-17
fix(output): adjustments for output text for different types of data

sanderblue:
Resolves: #333


sanderblue:
Closing and will revisit at a later time due to complexity.

PR #408, opened 2020-10-15
feat(newrelic): integrate with nrdiag

njvrzm:
This is a proof-of-concept for review! Do not merge :)

See `docs/nrdiag_integration.md` for motivation, design and what needs to be fleshed out.


ctrombley (commented):
Hey @njvrzm !  Thanks for the contribution!

I have a few thoughts about this approach.  Most of my comments here will address the differentiation of this approach from one where we compile the `nrdiag` logic in at CLI build time:

- First - I tested it out, it works, and it seemed relatively quick to implement.  That is a big win in incrementally delivering value, so I think we should move forward.
- This approach will not work for environments where internet access is restricted.  That may be an edge case but it's a possibility for some users.
- Users without chmod permissions will not be able to run the command successfully.  We've heard from the field that sometimes developers' machines are very locked down and they need IT support to install things.  Making sure that only needs to happen once (at CLI install time) is going to be somewhat important.  We've got a variety of ways that might happen - homebrew, standalone windows installers, and integrations with various other package managers.
- Controlling and minimizing the complicated conventions behind installation, versioning, packaging, distribution, and documentation was one of the goals of unifying behind a single CLI, and I worry that we are not solving that problem if we are just downloading and shelling out to another CLIs with its own conventions.
- I think it makes sense to log out what's happening with the download and install to the console
- I haven't tested this on Windows.

Some comments on the items in the "Next steps" section in the readme:
- Did you want to put auto updates in place?  Or just warn the user that the nrdiag command is out of date?  Fixed versions are important for reliability and repeatability, we see that for instance with Docker's advice to not pin to the `latest` tag in Dockerfiles.  Additionally, letting on that we are working with multiple versions and/or binaries becomes complicated for the user to understand.
- Passing arguments to the run command: if this is important, we will want the CLI help screens to show the proper documentation.  I'm not sure if this is possible without a tighter integration.  You mentioned that it doesn't feel reasonable to implement them in the `newrelic` cli and keep them in sync, but another option would be to shift the source of truth for command line args fully over to the `newrelic` CLI so we didn't have to maintain two implementations.
- For the `newrelic agent config lint` command, Cobra definitely supports command aliases.  We should be able to add this under the `nrdiag` sub-command and then also alias it under `agent` as well.
- I think cleaning up downloaded files is something we can just do automatically if file system permissions allow, do you agree?  I'm not sure if an option is warranted.

Overall I think we should run with this since its utility seems substantial compared to my concerns.  However I do worry that we run the risk of not addressing the concerns here because of other, higher priority projects. Let me know your thoughts - I am looking forward to taking this to the next step.  Thanks again!

njvrzm:
Great feedback, thank you @ctrombley! Notes below. We can talk more tomorrow and follow up as needed.

* This approach will not work for environments where internet access is restricted. That may be an edge case but it's a possibility for some users.

This did occur to me, and we should probably have some sort of handling in place to fail a bit more gracefully here. One nice thing about the general approach is that the user can put the nrdiag binary into place themselves if necessary.

* Users without chmod permissions...

Ah, this is a point I hadn't considered. Would love to do some brainstorming :)

* Controlling and minimizing the complicated conventions behind installation, versioning, packaging, distribution, and documentation was one of the goals of unifying behind a single CLI, and I worry that we are not solving that problem if we are just downloading and shelling out to another CLIs with its own conventions.

This is also my main concern with this approach. It's always a risk that "good enough for now" becomes the long-term solution. On the other hand, the design, planning and execution of a tighter integration seems like a pretty big project. (I _may_ soon be in a position to promise some help with that.)

* Do you want to put auto updates in place? ...

Now that you mention it, a warning and an update command would be a better idea. Just have to make sure to update in-place to avoid the chmod issue...

* Passing arguments to the run command: ...

This seems like maybe the thorniest issue here. It will almost certainly be necessary to provide at least _some_ command line arguments - proxy info, the `-y` flag, attachment key. Maybe we can only add the necessary flags, or maybe keeping them in both places wouldn't be too bad - as a short term solution. Something to discuss.

* I think cleaning up downloaded files is something we can just do automatically if file system permissions allow, do you agree? I'm not sure if an option is warranted.

I was thinking specifically of the binary, not the zip file, for example in case someone wanted to use nrdiag on a Docker container but not leave the ~12 meg file around. Also an edge case, maybe not worth addressing - and easy enough for the customer to do, if that's what they want.

PR #452, opened 2020-11-05
Ensure error is displayed and execution stops when a profile cannot be loaded

Julien4218:


ctrombley:
Closing this as the initialization process should not fail the invoked command if it fails.

PR #467, opened 2020-11-17
chore(build): update workflows to avoid deprecated add-path

zlesnr:
Without this change, the deprecation is preventing the jobs with add-path from
executing properly.  Here we update the setup and install of the workflows and
allow the PATH handling to be done by setup-go.


sanderblue (approved):
🦙 approves
🐑 it

PR #470, opened 2020-11-18
fix(install): use NEW_RELIC_CLI_VERSION within install script to avoid naming conflict

sanderblue:


sanderblue:
This actually turned out to be an issue with an old version of TLS on the openSUSE EC2 instance. Closing.

PR #472, opened 2020-11-20
chore(install): add helpers to detect linux os info

ctrombley:


ctrombley:
This can be done with `go-psutil` instead, closing this spike for now.

PR #475, opened 2020-11-20
fix(release): include the GOBIN path for the release

zlesnr:


ctrombley (approved):

zlesnr:
I don't think we need this any more.

PR #491, opened 2020-12-02
feat(install): prompt the user for accepting which of the discovered …

zlesnr:
…log files to watch

### DIFF
--- a/internal/install/install_test.go
+++ b/internal/install/install_test.go
@@ -3,11 +3,63 @@
 package install
 
 import (
+	"io/ioutil"
+	"os"
+	"path/filepath"
+	"sort"
 	"testing"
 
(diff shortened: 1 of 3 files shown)


internal/install/install_test.go
```diff
@@ -3,11 +3,63 @@
...
+	defer f2.Close()
+	f3, err := ioutil.TempFile(tmpDir, "*.nopelog")
+	require.NoError(t, err)
+	defer f3.Close()
+
+	r := recipeFile{}
+	r.MELTMatch.Logging = []logMatcher{
+		{
+			Name: "nginx",
+			File: filepath.Join(tmpDir, "*.log"),
```
zlesnr, line 41: `/tmp/*.log`

PR #494, opened 2020-12-03
chore(install): include glob file filterer

zlesnr:


internal/install/glob_file_filterer_test.go
```diff
@@ -0,0 +1,96 @@
+// +build unit
```
ctrombley, line 1: I would call these integration tests since we are interacting with the filesystem.

ctrombley (approved):
👨‍🍳

ctrombley (approved):

PR #508, opened 2020-12-10
chore(install): wire in the initial logging bits

zlesnr:


internal/install/go_task_recipe_executor.go
```diff
@@ -116,6 +120,20 @@ func setVarsFromSystemInfo(t *taskfile.Taskfile, m discoveryManifest) {
 	t.Vars.Merge(&v)
 }
 
+func setLiveVars(t *taskfile.Taskfile, vars map[string]interface{}) error {
```
ctrombley, line 123: Should we change this to `setStaticVars`?  `setProgrammaticVars`? Naming things is hard.

ctrombley (approved):

internal/install/go_task_recipe_executor.go
zlesnr: Good call.

ctrombley (approved):
🎉

PR #519, opened 2020-12-14
fix(build): skip go generate as part of build process

zlesnr:


ctrombley (approved):

Julien4218 (approved):
I'm approving, I dont know if you have any steps I could use to test it out?

PR #520, opened 2020-12-14
chore(install): add nerdstorageExecutionStatusReporter

ctrombley:


internal/install/execution_status.go
```diff
@@ -0,0 +1,89 @@
+package install
+
+import (
+	"time"
+
+	"github.com/google/uuid"
+)
+
+type executionStatus struct {
```
Julien4218, line 9: Not sure if there is a better name, but this name is confusing with the other type below executionStatusRecipe, particularly since this typoe does have a Recipe element in it, and the other type does not but has it in the name... confusing.

Julien4218 (approved):
That looks good beside perhaps a rename on one of the type.
cc @ctrombley

zlesnr (approved):

internal/install/execution_status.go
ctrombley: 👍

Julien4218 (approved):

PR #522, opened 2020-12-15
chore(install): assert profile region is US

Julien4218:


Julien4218:
Here is an example of output when run for an EU account:
FATAL Only the US region is supported at this time. region EU is not supported.

zlesnr (approved):

ctrombley (approved):

PR #535, opened 2020-12-18
fix(install): fixes for end to end flow

ctrombley:


internal/install/recipe_installer.go
```diff
@@ -100,11 +101,20 @@ func (i *recipeInstaller) install() error {
 		}
 	} else {
 		// Ask the recipe service for recommendations.
-		var err error
+		log.Debugln("Fetching recommendations.")
```
zlesnr, line 104: I think this line isn't needed.  The `fetchRecommendataions` might be spinning on this already.

zlesnr (dismissed):

internal/install/recipe_installer.go
ctrombley: 👍

zlesnr (approved):

PR #537, opened 2020-12-18
chore(install): suppress success message if one or more recipes fails to install

ctrombley:


zlesnr (approved):

zlesnr (approved):

zlesnr (approved):

Julien4218 (dismissed):
@ctrombley this is looking good.
The only addition I might suggest, we should fail the cli process if we see recipeErrorsEncountered is true.

internal/install/recipe_installer.go
```diff
@@ -160,21 +161,28 @@ func (i *RecipeInstaller) Install() error {
...
-		msg += fmt.Sprintf(`
-		https://one.newrelic.com/launcher/nrai.launcher?platform[accountId]=%d`, profile.AccountID)
+		profile := credentials.DefaultProfile()
+		if profile != nil {
+			msg += fmt.Sprintf(`
+			https://one.newrelic.com/launcher/nrai.launcher?platform[accountId]=%d`, profile.AccountID)
+		}
 	}
 
 	fmt.Println(msg)
```
ctrombley, line 188: `return errors.New(msg)` here?

Julien4218: Yeah, after Line 188 would be good.

Julien4218 (approved):

PR #544, opened 2020-12-21
fix(install): enable signal context notifications

zlesnr:


internal/install/command.go
```diff
@@ -28,6 +30,11 @@ var Command = &cobra.Command{
 	Short:  "Install New Relic.",
 	Hidden: true,
 	Run: func(cmd *cobra.Command, args []string) {
+		go func() {
+			<-utils.SignalCtx.Done()
+			os.Exit(1)
```
zlesnr, line 35: Maybe we want `os.Exit(130)` for "standards" adherence?  https://tldp.org/LDP/abs/html/exitcodes.html

PR #547, opened 2020-12-22
Prompt user to install integrations

ctrombley:


zlesnr (approved):

internal/install/ux/prompter.go
```diff
@@ -0,0 +1,5 @@
+package ux
+
+type Prompter interface {
+	PromptYesNo(msg string) (bool, error)
```
zlesnr, line 4: Do you also want to cover the case in `setVarsFromInput`?

PR #549, opened 2020-12-22
chore(install): clean up a few log messages

zlesnr:


internal/install/recipe_installer.go
```diff
@@ -408,7 +408,6 @@ func (i *RecipeInstaller) userAcceptsInstall(r types.Recipe) bool {
 }
 
 func (i *RecipeInstaller) fail(err error) error {
-	log.Error(err)
```
ctrombley, line 411: Good call, we are logging this one a level up in the stack as well.  Do we want to get rid of `fail()` altogether and just call `i.reportComplete()` in its place maybe?

ctrombley (approved):
LGTM with one comment
:shipit:

PR #563, opened 2020-12-28
refactor(install): avoid spinner while prompting

zlesnr:
Without this change, the UX for a user is less than ideal while prompting for
input vars on a given recipe.  Currently, the spinner for the task continues to
spin even while the process waits for input from the user.  To avoid confusing
the users, this work breaks the Execute() phase of a recipe execution into two
steps, Preapre() and Execute() to allow the process to gather all required input
up front and build the set of variables to be passed in to the Execute() phase.
This allows only the Execute() phase to show the spinner, while the Prepare()
phase does not.


devfreddy (approved):
👍

PR #566, opened 2020-12-29
Display name again

zlesnr:


internal/install/types/types.go
```diff
@@ -104,86 +104,90 @@ var OpenInstallationTargetTypeTypes = struct {
...
+	// Friendly name of the integration
+	DisplayName string `json:"displayName,omitempty"`
 	// The full contents of the recipe file (yaml)
 	File string `json:"file"`
 	// The ID
-	ID string `json:"id"`
+	ID string `json:"id,omitempty"`
 	// List of variables to prompt for input from the user
 	InputVars []OpenInstallationRecipeInputVariable `json:"inputVars"`
+	// Go-task's taskfile definiton (see https://taskfile.dev/#/usage)
```
devfreddy, line 142: Not important but maybe to satisfy your linter, correct the spelling of `definiton` to `definition`

devfreddy (dismissed):
👍 with a low-priority comment on a spelling thing

internal/install/types/types.go
devfreddy: Also, apologies, b/c this almost definitely came from you copy/pasting my spelling mistake 😂

zlesnr: This code is all generated, so what is stored in the schema is what gets written here.  I'll correct it in place now.

devfreddy (approved):
:shipit:

PR #568, opened 2020-12-30
chore(install): include verbose and trace flags for loglevel

zlesnr:


internal/install/ux/spinner.go
```diff
@@ -42,7 +42,7 @@ func (s *Spinner) Start(msg string) {
 
 func (s *Spinner) Stop() {
 	// Only stop the spinner of the log level is info or below.
-	if log.GetLevel() <= log.InfoLevel {
+	if log.IsLevelEnabled(log.InfoLevel) {
```
ctrombley, line 45: 👍

internal/install/ux/spinner.go
```diff
@@ -31,7 +31,7 @@ func NewSpinner() *Spinner {
 
 func (s *Spinner) Start(msg string) {
 	// Only start the spinner of the log level is info or below.
-	if log.GetLevel() > log.InfoLevel {
+	if log.IsLevelEnabled(log.DebugLevel) {
```
ctrombley, line 34: 👍

internal/install/command.go
```diff
@@ -73,4 +81,6 @@ func init() {
 	Command.Flags().BoolVarP(&skipIntegrations, "skipIntegrations", "r", false, "skips installation of recommended New Relic integrations")
 	Command.Flags().BoolVarP(&skipLoggingInstall, "skipLoggingInstall", "l", false, "skips installation of New Relic Logging")
 	Command.Flags().BoolVarP(&testMode, "testMode", "t", false, "fakes operations for UX testing")
+	Command.Flags().BoolVar(&verbose, "verbose", false, "increase log level verbosity")
```
ctrombley, line 84: Consider long flag `debug`, "debug level logging" for description and short flag `v`?

internal/install/command.go
```diff
@@ -73,4 +81,6 @@ func init() {
...
+	Command.Flags().BoolVar(&trace, "trace", false, "trace level logging")
```
ctrombley, line 85: Consider short flag `vv`?

internal/install/command.go
zlesnr: `-v` is taken for verbose currently on the root command, which we could change, but `--debug` sounds find to me too.

internal/install/command.go
zlesnr: I did and then read https://github.com/spf13/cobra/issues/679

ctrombley (approved):

PR #575, opened 2021-01-04
chore(install): skip prompting on recipes loaded from file path

zlesnr:


Julien4218 (dismissed):

internal/install/recipe_installer.go
```diff
@@ -200,7 +210,7 @@ func (i *RecipeInstaller) installRecipesWithPrompts(m *types.DiscoveryManifest,
 		var err error
 
 		// Skip prompting the user if the recipe has been asked for directly.
-		if namedArgument(r.Name) {
+		if namedArgument(r.Name) || pathedArgument(r.FilePath) {
```
ctrombley, line 213: can't we just use `i.RecipesProvided()` here and avoid the inline func?

ctrombley (approved):
🌮

PR #577, opened 2021-01-05
feat(install): add -y flag

zlesnr:
Included here is a new command flag which will allow the user to accept all
prompts using default values if none are set in the environment to use instead.

### DIFF
--- a/internal/install/execution/go_task_recipe_executor.go
+++ b/internal/install/execution/go_task_recipe_executor.go
@@ -195,7 +195,7 @@ func varsFromInput(inputVars []recipes.VariableConfig, assumeYes bool) (types.Re
 		if envValue == "" {
 			if assumeYes {
 				log.Debugf("required env var %s not found, using default value", envConfig.Name)
-				envValue = envConfig.Default
+				vars[envConfig.Name] = envConfig.Default
 			} else {
 				log.Debugf("required env var %s not found, prompting for input", envConfig.Name)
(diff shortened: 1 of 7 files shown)


internal/install/execution/go_task_recipe_executor.go
```diff
@@ -186,37 +186,43 @@ func varsFromRecipe(r types.Recipe) (types.RecipeVars, error) {
...
-			if envConfig.Default != "" {
-				prompt.Default = envConfig.Default
-			}
-
-			result, err := prompt.Run()
-			if err != nil {
-				return types.RecipeVars{}, fmt.Errorf("prompt failed: %s", err)
+		if envValue == "" {
+			if assumeYes {
+				log.Debugf("required env var %s not found, using default value", envConfig.Name)
```
Julien4218, line 197: @zlesnr Is there any issue when the assumeYes is used but a recipe requires an input to be entered and no default is provided?
I believe in this case some sort of error should raised (maybe this is done somewhere else), or alternatively the question should be prompted (perhaps bad given the input said not to?).

zlesnr: Good callout.  I'll draft a workaround.

zlesnr: I've updated the code to check for empty default value and return an error if none were found.  How's that?

sanderblue (dismissed):
🐑 it

sanderblue (approved):
Re 🐑 it

PR #578, opened 2021-01-05
chore(install): refactor varsFromInput to reduce conditional nesting

ctrombley:


internal/install/execution/go_task_recipe_executor.go
```diff
@@ -190,47 +190,53 @@ func varsFromInput(inputVars []recipes.VariableConfig, assumeYes bool) (types.Re
...
+		}
 
-				log.Debugf("required env var %s not found, using default value", envConfig.Name)
-				vars[envConfig.Name] = envConfig.Default
-			} else {
-				log.Debugf("required env var %s not found, prompting for input", envConfig.Name)
-				msg := fmt.Sprintf("value for %s required", envConfig.Name)
+		if assumeYes {
+			if envConfig.Default == "" {
+				return vars, fmt.Errorf("no default value for environment variable %s and none provided", envConfig.Name)
```
zlesnr, line 203: Do we want to match the other return of `types.RecipeVars{}` rather than the incomplete set of vars here?

zlesnr (dismissed):

internal/install/execution/go_task_recipe_executor.go
ctrombley: 👍

zlesnr (approved):

PR #585, opened 2021-01-07
fix(logging): set client logger level

zlesnr:
Without this change, the client is not aware of the user's desires, and the log
level is confined to the logged messages in the CLI code.  Here we use a new
method on the client to set the log level when debug or trace are requested by
the user.


zlesnr:
Requires: https://github.com/newrelic/newrelic-client-go/pull/586

sanderblue (approved):

PR #588, opened 2021-01-07
chore(install): allow non us regions, and add more debug statements. …

Julien4218:
…Fix issue when recipe is not executed with -c

### DIFF
--- a/internal/install/recipe_installer.go
+++ b/internal/install/recipe_installer.go
@@ -99,8 +99,7 @@ func (i *RecipeInstaller) Install() error {
 				log.Debugln(fmt.Sprintf("Error while building recipe from path, detail:%s.", err))
 				return i.fail(err)
 			}
-
-			log.Debugln("Found recipe from path %s.", n)
+			log.Debugln(fmt.Sprintf("Found recipe from path %s.", n))
 			recipes = append(recipes, *recipe)
 		}
 	} else if i.RecipeNamesProvided() {
@@ -156,7 +155,7 @@ func (i *RecipeInstaller) Install() error {
 			return i.fail(err)
 		}
 		log.Debugf("Done installing infrastructure agent.")
-	}	else {
+	} else {
 		log.Debugf("Skipping installing infrastructure agent")
 	}
 
@@ -168,7 +167,7 @@ func (i *RecipeInstaller) Install() error {
 			return i.fail(err)
 		}
 		log.Debugf("Done installing logging.")
-	}	else {
+	} else {
 		log.Debugf("Skipping installing logging")
 	}
 
@@ -180,7 +179,7 @@ func (i *RecipeInstaller) Install() error {
 			return err
 		}
 		log.Debugf("Done installing integrations.")
-	}	else {
+	} else {
 		log.Debugf("Skipping installing integrations")
 	}
 
@@ -214,7 +213,7 @@ func (i *RecipeInstaller) installRecipesWithPrompts(m *types.DiscoveryManifest, 
 		// The infra and logging install have their own install methods.  In the
 		// case where the recommendations come back with either of these recipes,
 		// we skip here to avoid duplicate installation.
-		if i.RecipesProvided() == false{
+		if !i.RecipesProvided() {
 			if r.Name == infraAgentRecipeName || r.Name == loggingRecipeName {
 				log.Debugf(fmt.Sprintf("Skipping recipe %s with prompts, matching either infra agent name %s or logging recipe name %s.", r.Name, infraAgentRecipeName, loggingRecipeName))
 				continue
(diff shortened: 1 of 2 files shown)


internal/install/recipe_installer.go
```diff
@@ -123,36 +130,57 @@ func (i *RecipeInstaller) Install() error {
...
 
 		for _, r := range recipes {
-			log.Debugf("Found available integration %s.", r.Name)
+			log.Debugf(fmt.Sprintf("Found available integration %s.", r.Name))
 		}
 
 		i.reportRecipesAvailable(recipes)
 	}
 
+	log.Debugf(fmt.Sprintf("AssumeYes:%t.", i.AssumeYes))
```
ctrombley, line 139: This multi-statement debug block can also be achieved with `log.Debugf(%+v\n.", i.InstallerContext)`, but with slightly harder-to-read format.

internal/install/recipe_installer.go
```diff
@@ -178,13 +206,18 @@ func (i *RecipeInstaller) Install() error {
 
 func (i *RecipeInstaller) installRecipesWithPrompts(m *types.DiscoveryManifest, recipes []types.Recipe, entityGUID string) (bool, error) {
 	errorsEncountered := false
+	log.Debugf("Installing recipes with prompts...")
 
 	for _, r := range recipes {
+		log.Debugf(fmt.Sprintf("Installing recipe %s with prompts...", r.Name))
```
ctrombley, line 212: `fmt.Sprintf` is unnecessary from within `log.Debugf` - you can use it the same way you would `Sprintf`.

internal/install/recipe_installer.go
```diff
@@ -194,29 +227,36 @@ func (i *RecipeInstaller) installRecipesWithPrompts(m *types.DiscoveryManifest,
 		if i.RecipesProvided() || i.AssumeYes {
 			ok = true
 		} else {
+			log.Debugf("Checking user accepts install...")
 			ok, err = i.userAcceptsInstall(r)
 			if err != nil {
+				log.Debugf(fmt.Sprintf("Done installing recipes with prompts, exception:%s", err))
```
ctrombley, line 233: fmt.Sprintf is unnecessary from within log.Debugf - you can use it the same way you would Sprintf.

internal/install/recipe_installer.go
```diff
@@ -194,29 +227,36 @@ func (i *RecipeInstaller) installRecipesWithPrompts(m *types.DiscoveryManifest,
...
 				return true, err
 			}
+			log.Debugf(fmt.Sprintf("Done checking user accepts install ok:%t", ok))
 		}
 
 		if !ok {
-			log.Debugf("skipping recipe %s.", r.Name)
+			log.Debugf(fmt.Sprintf("skipping not ok recipe %s.", r.Name))
```
ctrombley, line 240: fmt.Sprintf is unnecessary from within log.Debugf - you can use it the same way you would Sprintf.

internal/install/recipe_installer.go
```diff
@@ -194,29 +227,36 @@ func (i *RecipeInstaller) installRecipesWithPrompts(m *types.DiscoveryManifest,
 		if i.RecipesProvided() || i.AssumeYes {
 			ok = true
 		} else {
+			log.Debugf("Checking user accepts install...")
 			ok, err = i.userAcceptsInstall(r)
 			if err != nil {
+				log.Debugf(fmt.Sprintf("Done installing recipes with prompts, exception:%s", err))
 				return true, err
 			}
+			log.Debugf(fmt.Sprintf("Done checking user accepts install ok:%t", ok))
```
ctrombley, line 236: fmt.Sprintf is unnecessary from within log.Debugf - you can use it the same way you would Sprintf.

internal/install/recipe_installer.go
Julien4218: Nice, I was wishing for something like that, I'll update. I don't mind that it is harder to read in debug. It's also more maintainable if more fields are added in the future.

internal/install/recipe_installer.go
Julien4218: I was hoping for that, but it's not what I'm noticing when it runs with just the log.Debugf( ). In the output, I see the %s in the message, and the parameter added after the entire string.

internal/install/recipe_installer.go
Julien4218: Ho my bad, I've mistaken log.DebugLn with log.Debugf.
Do I need to explicit do a line return if I use log.Debugf ?

internal/install/recipe_installer.go
Julien4218: Ho my bad, I've mistaken log.DebugLn with log.Debugf.
Do I need to explicit do a line return if I use log.Debugf ?

ctrombley (approved):
:shipit:

internal/install/recipe_installer.go
```diff
@@ -92,29 +92,36 @@ func (i *RecipeInstaller) Install() error {
 	if i.RecipePathsProvided() {
 		// Load the recipes from the provided file names.
 		for _, n := range i.RecipePaths {
+			log.Debugln(fmt.Sprintf("Attempting to match recipePath %s.", n))
```
zlesnr, line 95: `Debugf` is a good way to combine this into smaller code.  `log.Debugf("Attempting to match recipePath %s.", n)` is a little less to look at

zlesnr (commented):
Thanks for this.

PR #594, opened 2021-01-08
chore(install): report infra and logging recipes available

ctrombley:


internal/install/recipe_installer.go
```diff
@@ -130,45 +140,40 @@ func (i *RecipeInstaller) Install() error {
...
 
-	log.Debugf(fmt.Sprintf("AssumeYes:%t.", i.AssumeYes))
-	log.Debugf(fmt.Sprintf("RecipeNames:%s.", i.RecipeNames))
-	log.Debugf(fmt.Sprintf("RecipePaths:%s.", i.RecipePaths))
-	log.Debugf(fmt.Sprintf("SkipDiscovery:%t.", i.SkipDiscovery))
-	log.Debugf(fmt.Sprintf("SkipInfraInstall:%t.", i.SkipInfraInstall))
-	log.Debugf(fmt.Sprintf("SkipLoggingInstall:%t.", i.SkipLoggingInstall))
-	log.Debugf(fmt.Sprintf("SkipIntegrations:%t.", i.SkipIntegrations))
-	log.Debugf(fmt.Sprintf("RecipesProvided:%t.", i.RecipesProvided()))
+	log.Debugf("InstallerContext: %+v\n", i.InstallerContext)
```
Julien4218, line 149: Can we keep the debug statement for the output of the method i.RecipesProvided() ?

internal/install/recipe_installer.go
```diff
@@ -130,45 +140,40 @@ func (i *RecipeInstaller) Install() error {
 		}
 
 		for _, r := range recipes {
-			log.Debugf(fmt.Sprintf("Found available integration %s.", r.Name))
+			log.Debugf("Found available integration %s.", r.Name)
```
Julien4218, line 143: Do we need a line feed at the end?

internal/install/recipe_installer.go
```diff
@@ -88,6 +88,16 @@ func (i *RecipeInstaller) Install() error {
 		return i.fail(err)
 	}
 
+	infraAgentRecipe, err := i.fetchRecipeAndReportAvailable(m, infraAgentRecipeName)
```
Julien4218, line 91: Is it relevant to fetch the infra and logging recipe when the `-c` flag is used?

internal/install/recipe_installer.go
zlesnr: We shouldn't.  The `log` statement should do that for us.  If it were a `fmt.Sprintf`, then we would.

zlesnr (approved):
🌯

internal/install/recipe_installer.go
ctrombley: 👍

internal/install/recipe_installer.go
ctrombley: I think it is, since we want to show it as "Available" yet "Skipped" in the UI.

Julien4218: Sorry to insist @ctrombley but I don't think it's valid to fetch the infraAgentRecipe and loggingRecipe at this level.
For one, those are NOT used when a recipe is provided explicitly with the -c option, which is NOT the workflow of what a user would do from the NR UI.
Second, if we look at the usages of those 2 recipe objects, they are only used later down in the ShouldInstallInfraAgent( ) and ShouldInstallLogging( ) method, and symmetrically those are not used for the ShouldInstallIntegrations( ) path which is the way the recipe path provided execute (-c option).
Consequently, could we just move the fetch of infra/logging closer to the scope where they are used in their respective ShouldInstallXxx( ) methods?

Julien4218 (dismissed):

Julien4218:
Looking good!

Julien4218 (approved):

PR #598, opened 2021-01-11
fix(install): avoid newline in prompt message

zlesnr:


zlesnr:
Related to https://github.com/manifoldco/promptui/issues/170

ctrombley (approved):
⚡

PR #624, opened 2021-01-20
chore(install): add validation and filtering for platform and platform family

ctrombley:
This PR uses reflection over generated types to ensure that we are not sending invalid values for `platform` or `platformFamily` when retrieving recommendations from the recipe service.

cc @jbeveland27  @michaelkro

### DIFF
--- a/internal/install/discovery/psutil_discoverer.go
+++ b/internal/install/discovery/psutil_discoverer.go
@@ -3,6 +3,7 @@ package discovery
 import (
 	"context"
 	"fmt"
+	"reflect"
 
 	"github.com/shirou/gopsutil/host"
 	"github.com/shirou/gopsutil/process"
@@ -39,6 +40,8 @@ func (p *PSUtilDiscoverer) Discover(ctx context.Context) (*types.DiscoveryManife
 		PlatformVersion: i.PlatformVersion,
 	}
 
+	m = filterValues(m)
+
 	pids, err := process.PidsWithContext(ctx)
 	if err != nil {
 		return nil, fmt.Errorf("cannot retrieve processes: %s", err)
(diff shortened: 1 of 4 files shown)


internal/install/discovery/psutil_discoverer.go
```diff
@@ -67,3 +70,41 @@ func (p *PSUtilDiscoverer) Discover(ctx context.Context) (*types.DiscoveryManife
...
+
+	return m
+}
+
+func isValidOpenInstallationPlatform(platform string) bool {
+	s := reflect.ValueOf(&types.OpenInstallationPlatformTypes).Elem()
+
+	for i := 0; i < s.NumField(); i++ {
+		v := s.Field(i).Interface().(types.OpenInstallationPlatform)
+		if string(v) == platform {
```
Julien4218, line 91: @ctrombley Is this a case insensitive comparison? I believe we might want that.

internal/install/discovery/psutil_discoverer.go
```diff
@@ -67,3 +70,41 @@ func (p *PSUtilDiscoverer) Discover(ctx context.Context) (*types.DiscoveryManife
...
+
+	return false
+}
+
+func isValidOpenInstallationPlatformFamily(platformFamily string) bool {
+	s := reflect.ValueOf(&types.OpenInstallationPlatformFamilyTypes).Elem()
+
+	for i := 0; i < s.NumField(); i++ {
+		v := s.Field(i).Interface().(types.OpenInstallationPlatformFamily)
+		if string(v) == platformFamily {
```
Julien4218, line 104: @ctrombley Is this a case insensitive comparison? I believe we might want that.

internal/install/discovery/psutil_discoverer.go
```diff
@@ -67,3 +70,41 @@ func (p *PSUtilDiscoverer) Discover(ctx context.Context) (*types.DiscoveryManife
 
 	return &m, nil
 }
+
+func filterValues(m types.DiscoveryManifest) types.DiscoveryManifest {
+	if !isValidOpenInstallationPlatform(m.Platform) {
```
Julien4218, line 75: Should the default be nil instead of empty string? I wonder if we'll find some usages for empty strings later on.

Julien4218 (approved):

Julien4218:
@ctrombley a side note, could we add some debug logging to the QueryWithResponseAndContext( ) calls for what goes out and back, this way we can troubleshoot more easily when things don't work?

internal/install/discovery/psutil_discoverer.go
ctrombley: I believe that graphQL is case-sensitive when checking enums.

internal/install/discovery/psutil_discoverer.go
ctrombley: This would require updating the type to a string pointer.  Aside from the fact that the empty string would currently also be rejected by graphQL, idiomatic Go says to use normal strings unless nil is necessary. Normal strings are easier and safer to use, without the need to check that`*string` has a value before dereferencing.

internal/install/discovery/psutil_discoverer.go
ctrombley: I believe that graphQL is case-sensitive when checking enums.

internal/install/discovery/psutil_discoverer.go
Julien4218: No problem, thanks for the details.

PR #636, opened 2021-01-21
fix(install): avoid duplicate results when fetching recommendations

zlesnr:
Without this change, the duplicate fix in 9ee4a93 is too aggressive in removing
the duplicates.  In the case where we fetch recipes, we want all results so that
we are able to effectively determine all of the process filters regardless of
the name of the recipe.  Here we shift that duplicate avoidance to the
FetchRecommendations call instead, since we only want to remove the duplicates
in the case the recipe service sends back a duplicate when we are looking for
recommendations.

### DIFF
--- a/internal/install/recipes/service_recipe_fetcher.go
+++ b/internal/install/recipes/service_recipe_fetcher.go
@@ -77,7 +77,29 @@ func (f *ServiceRecipeFetcher) FetchRecommendations(ctx context.Context, manifes
 		return nil, err
 	}
 
-	return resp.Docs.OpenInstallation.Recommendations.ToRecipes(), nil
+	allRecipes := resp.Docs.OpenInstallation.Recommendations.ToRecipes()
+
+	r := []types.Recipe{}
+
+	recipeIncluded := func(recipe types.Recipe, recipes []types.Recipe) bool {
+		for _, r := range recipes {
+			if recipe.Name == r.Name {
+				return true
+			}
+		}
+
+		return false
+	}
+
+	for _, recipe := range allRecipes {
+		if recipeIncluded(recipe, r) {
+			continue
+		}
+
+		r = append(r, recipe)
+	}
+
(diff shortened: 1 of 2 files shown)


internal/install/recipes/service_recipe_fetcher.go
```diff
@@ -77,7 +77,29 @@ func (f *ServiceRecipeFetcher) FetchRecommendations(ctx context.Context, manifes
 		return nil, err
 	}
 
-	return resp.Docs.OpenInstallation.Recommendations.ToRecipes(), nil
+	allRecipes := resp.Docs.OpenInstallation.Recommendations.ToRecipes()
```
Julien4218, line 80: Do we need this dedup logic?
Normally the searchRecipe and process detection should match to subset of recipes, and when the recommendations are fetched it should be leveraging the process match to only pick 1 recipe for a given name.

Maybe there is another change coming, but another key part is to add the os/platform criteria when we search for recipes. For example we should only return recipes that are meant for the current os and/or linux version. We wouldn't want to search and returns recipes that are not meant to be installed on the current host os/platform.

zlesnr: I believe this was mostly a guard against the service sending back two recipes with the same name.  This was the case in some circumstances that caused the recipes to be installed more than once in a single run.

I don't think this affects the process filtering currently, but are you suggesting an additional change to the process filter?

Julien4218: Got it, maybe this is needed I'm unsure. I dont quite like that we arbitrarily pick one, maybe we picked the wrong one?

Yeah, I think we need another change that might prevent the case of having duplicate named recipes coming back.
For example, if I'm running on an aws linux2 os/platform, I shouldn't get back a searched recipe that targets debian, or even more so a recipe targeting windows.

PR #637, opened 2021-01-21
chore(install): Update process matching to use whole command line

Julien4218:


internal/install/types/discovery_manifest.go
```diff
@@ -2,23 +2,30 @@ package types
...
-	PlatformVersion string           `json:"platformVersion"`
-	Processes       []GenericProcess `json:"processes"`
+	Hostname        string            `json:"hostname"`
+	KernelArch      string            `json:"kernelArch"`
+	KernelVersion   string            `json:"kernelVersion"`
+	OS              string            `json:"os"`
+	Platform        string            `json:"platform"`
+	PlatformFamily  string            `json:"platformFamily"`
+	PlatformVersion string            `json:"platformVersion"`
+	Processes       []ProcessInfoWrap `json:"processes"`
```
ctrombley, line 12: Consider renaming to `MatchedProcesses`?

internal/install/types/discovery_manifest.go
```diff
@@ -2,23 +2,30 @@ package types
...
 }
 
 // GenericProcess is an abstracted representation of a process.
 type GenericProcess interface {
 	Name() (string, error)
+	Cmdline() (string, error)
 	PID() int32
 }
 
+type ProcessInfoWrap struct {
```
ctrombley, line 22: Consider renaming to `MatchedProcess`?

internal/install/types/discovery_manifest.go
```diff
@@ -2,23 +2,30 @@ package types
...
+	Info            string
```
ctrombley, line 23: Consider renaming to `CmdLine` or `Command`?

internal/install/types/discovery_manifest.go
```diff
@@ -2,23 +2,30 @@ package types
...
+	Process         GenericProcess
+	MatchingPattern string
+}
+
 // AddProcess adds a discovered process to the underlying manifest.
-func (d *DiscoveryManifest) AddProcess(p GenericProcess) {
+func (d *DiscoveryManifest) AddProcess(p ProcessInfoWrap) {
```
ctrombley, line 29: Consider renaming to `AddMatchedProcess`?

internal/install/discovery/regex_process_filterer.go
```diff
@@ -23,45 +23,68 @@ func NewRegexProcessFilterer(r recipes.RecipeFetcher) *RegexProcessFilterer {
...
 	}
 
 	return false
 }
+
+func getProcessesInfo(processes []types.GenericProcess) []types.ProcessInfoWrap {
+	processesInfo := []types.ProcessInfoWrap{}
+	for _, p := range processes {
+		cmdLine, err := p.Cmdline()
+		if err == nil {
```
ctrombley, line 79: nit: invert `if` here and use a `continue` to avoid nesting, and log a message at Debug level?

internal/install/discovery/regex_process_filterer.go
```diff
@@ -23,45 +23,68 @@ func NewRegexProcessFilterer(r recipes.RecipeFetcher) *RegexProcessFilterer {
 	return &f
 }
 
-func (f *RegexProcessFilterer) filter(ctx context.Context, processes []types.GenericProcess) ([]types.GenericProcess, error) {
+func (f *RegexProcessFilterer) filter(ctx context.Context, processes []types.GenericProcess) ([]types.ProcessInfoWrap, error) {
+	processesInfo := getProcessesInfo(processes)
```
ctrombley, line 27: Consider renaming `processesInfo` to `runningProcesses`, and returning a slice of `GenericProcess` from the method here?  We can resolve the `CmdLine()` value down below when we perform the matching.

internal/install/discovery/regex_process_filterer.go
```diff
@@ -23,45 +23,68 @@ func NewRegexProcessFilterer(r recipes.RecipeFetcher) *RegexProcessFilterer {
...
+			processInfo.MatchingPattern = pattern
+			log.Debugf("Process matching pattern %s with %s for recipe %s.", pattern, processInfo.Info, r.DisplayName)
 			return matched
 		}
 	}
 
 	return false
 }
+
+func getProcessesInfo(processes []types.GenericProcess) []types.ProcessInfoWrap {
```
ctrombley, line 75: Consider renaming to `getRunningProcesses` and returning a slice of `GenericProcess` from the method here?  We can resolve the `CmdLine()` values when we perform the matching, we don't have to do it here.

internal/install/discovery/regex_process_filterer.go
```diff
@@ -23,45 +23,68 @@ func NewRegexProcessFilterer(r recipes.RecipeFetcher) *RegexProcessFilterer {
...
+	processesInfo := []types.ProcessInfoWrap{}
+	for _, p := range processes {
+		cmdLine, err := p.Cmdline()
+		if err == nil {
+			if len(cmdLine) > 0 {
```
ctrombley, line 80: nit: cmdLine != ""  is more idiomatic for checking zero value of a string

internal/install/discovery/regex_process_filterer.go
```diff
@@ -23,45 +23,68 @@ func NewRegexProcessFilterer(r recipes.RecipeFetcher) *RegexProcessFilterer {
...
 			matches = append(matches, p)
 		}
 	}
 
+	log.Debugf("Filtering recipes with processes done, found %d matches.", len(matches))
 	return matches, nil
 }
 
-func match(r types.Recipe, process types.GenericProcess) bool {
+func match(r types.Recipe, processInfo *types.ProcessInfoWrap) bool {
```
ctrombley, line 57: Consider passing a GenericProcess in, and then returning a pointer to a `ProcessInfoWrap` (or whatever else we end up calling it) to avoid the side effects on the `processInfo` struct.

ctrombley (changes requested):

internal/install/discovery/regex_process_filterer.go
```diff
@@ -23,45 +23,68 @@ func NewRegexProcessFilterer(r recipes.RecipeFetcher) *RegexProcessFilterer {
 	return &f
 }
 
-func (f *RegexProcessFilterer) filter(ctx context.Context, processes []types.GenericProcess) ([]types.GenericProcess, error) {
+func (f *RegexProcessFilterer) filter(ctx context.Context, processes []types.GenericProcess) ([]types.ProcessInfoWrap, error) {
+	processesInfo := getProcessesInfo(processes)
+	log.Debugf("Filtering recipes with %d processes...", len(processesInfo))
+	for _, p := range processesInfo {
+		log.Debugf("Match using processInfo: %s", p.Info)
```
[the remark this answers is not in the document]
Julien4218: Done

internal/install/types/discovery_manifest.go
Julien4218: Done

internal/install/types/discovery_manifest.go
Julien4218: Done

internal/install/discovery/regex_process_filterer.go
Julien4218: Done, except for log message, I'm not sure there was something relevant to output.

internal/install/discovery/regex_process_filterer.go
Julien4218: done. FYI I've googled that very particular topic and it seems people were split among those 2 semantic. I don't particularly care, so I've switched it.

Julien4218:
@ctrombley I've made most of the changes. 
The only part I've left dealt with the lifecyle of all the MatchedProcess, slice and side effects. There is a bit of a wrinkle with the call to Cmdline( ) which is done early for debug/trace purpose, and also the wrinkle that this method call may fail for whatever reason.

ctrombley (approved):

PR #638, opened 2021-01-22
fix(install): ensure all scope writes during complete reporting

zlesnr:
Without this change, only the user scope is updated, even when we know the
entities that should be updated as well.  Here we ensure that the entity scope
is included as well when we report complete.


zlesnr:
This is probably not required.  Chatter is happening elsewhere.

PR #643, opened 2021-01-25
fix(install): capture guid from validation when alternate format

zlesnr:
Without this change, the GUID is not returned from a logs validation query due
to the different format.  The standard case seems to use entityGuid, while logs
uses entity.guids.  Here we ensure we capture both cases for the return.


sanderblue (approved):
🐑 it

PR #649, opened 2021-01-29
Ux fixes

zlesnr:


internal/install/recipe_installer.go
```diff
@@ -87,6 +87,18 @@ func (i *RecipeInstaller) Install() error {
 	`)
 	fmt.Println()
 
+	log.Tracef("InstallerContext: %+v", i.InstallerContext)
+	log.WithFields(log.Fields{
```
ctrombley, line 91: 👍

internal/install/recipe_installer.go
```diff
@@ -165,45 +183,43 @@ func (i *RecipeInstaller) Install() error {
...
 
 		if i.SkipInfraInstall {
 			i.status.ReportRecipeSkipped(execution.RecipeStatusEvent{Recipe: *infraAgentRecipe})
 		} else {
-			log.Debugf("Installing infrastructure agent")
-			entityGUID, err = i.executeAndValidateWithProgress(m, infraAgentRecipe)
-			if err != nil {
-				log.Error(i.failMessage(infraAgentRecipeName))
-				return i.fail(err)
+			if i.ShouldInstallInfraAgent() {
```
ctrombley, line 192: Is this necessary within the current block?  `SkipInfraInstall` has already evaluated to `false` here.  It's a bit confusing that this needs to be checked again.  Maybe a rename of the method would help with clarity?

internal/install/recipe_installer.go
```diff
@@ -165,45 +183,43 @@ func (i *RecipeInstaller) Install() error {
...
+				}
+				log.Debugf("Done installing infrastructure agent.")
 			}
-			log.Debugf("Done installing infrastructure agent.")
 		}
 
 		if i.SkipLoggingInstall {
 			i.status.ReportRecipeSkipped(execution.RecipeStatusEvent{Recipe: *loggingRecipe})
 		} else {
+			if i.ShouldInstallLogging() {
```
ctrombley, line 206: Same comment as above here.

internal/install/recipe_installer.go
zlesnr: Yeah, I see how that's a little confusing, though I think the logic is accurate in the case where a recipe is installed.  The `Should` method checks if there were recipes provided too, as well as the skip flag.  So in the case of the skip flag, we want to report skipped, but in the case a recipe (name or path) was provided, I don't think we do.  I suppose if it didn't matter, the `if` at the top of the block could just check `ShouldInstall...()` and then the `else` would report the skip.  What do you think?

ctrombley: LGTM.

ctrombley (approved):

Issue #623, opened 2021-01-19
NR CLI Translate base64 encoded urls command


jpvajda:
In the Seldon era, the NR URLs produced were human readable. With NR1 all of the urls are base64 encoded JSON strings, these are not human readable. There has been advancements to make them more human readable, but there are still some use cases where it is helpful to know the Application ID or Condition ID, information that is not readily available from the UI.

 it takes an existing NR1 URL, and parses it from base64 to a human readable JSON format. It even handles the situations where the encoded string contains an encoded entity ID, as it is quite often this entity ID information that we require.

### TO DO
Add a command that implements the behavior of the [internal tool](https://source.datanerd.us/nfoley/base64decoder) that was built to take care of this use case and add the  functionality to the NR CLI.

### Command 

`translate`  

-h, --help  
 --format string   output text format [JSON, Text, YAML] (default "JSON")
 --plain           output compact text

Example URL (works for any NR1 URL)

```
    https://one.newrelic.com/launcher/nr1-core.explorer?pane=eyJuZXJkbGV0SWQiOiJhcG0tbmVyZGxldHMub3ZlcnZpZXciLCJpc092ZXJ2aWV3Ijp0cnVlLCJlbnRpdHlJZCI6Ik1UWXdOamcyTW54QlVFMThRVkJRVEVsRFFWUkpUMDU4TlRBeE1UWXpORGcifQ==&sidebars[0]=eyJuZXJkbGV0SWQiOiJucjEtY29yZS5hY3Rpb25zIiwiZW50aXR5SWQiOiJNVFl3TmpnMk1ueEJVRTE4UVZCUVRFbERRVlJKVDA1OE5UQXhNVFl6TkRnIiwic2VsZWN0ZWROZXJkbGV0Ijp7Im5lcmRsZXRJZCI6ImFwbS1uZXJkbGV0cy5vdmVydmlldyIsImlzT3ZlcnZpZXciOnRydWV9fQ==&platform[timeRange][duration]=1800000&platform[$isFallbackTimeRange]=true

```

current Output in internal tool

```

    ------------------------
    485171|APM|APPLICATION|676227677
    ------------------------
    {"nerdletId":"apm-nerdlets.overview","isOverview":true,"entityId":"NDg1MTcxfEFQTXxBUFBMSUNBVElPTnw2NzYyMjc2Nzc"}

```

desired Output 

```
[...]

PR #653, opened 2021-02-01
chore(install): NR CLI Translate base64 encoded urls command #623

GlenOFoghlu:
Created this PR to create Decode CLI operation -> https://github.com/newrelic/newrelic-cli/issues/623


cmd/newrelic/command_decode.go
```diff
@@ -0,0 +1,61 @@
...
+		relicString = u.Query().Get("pane")
+		decoded, err := base64.StdEncoding.DecodeString(relicString)
+
+		if err != nil {
+			fmt.Println("decode error:", err)
+		}
+
+		decodedURL := string(decoded)
+		fmt.Println("\n**** Decoded URL **** ")
+		fmt.Println(decodedURL)
```
sanderblue, line 32: Can we delete these two lines of output? This looks like debugging output 🙂

cmd/newrelic/command_decode.go
```diff
@@ -0,0 +1,61 @@
...
+)
+
+// Command represents the decode command.
+var cmdDecode = &cobra.Command{
+	Use:     "decode",
+	Short:   "Decodes NR1 URL Strings ",
+	Example: `newrelic decode https://one.newrelic.com/launcher/nr1-core.home?pane=eyJuZXJkbGV0SWQiOiJucjEtY29yZS5ob21lLXNjcmVlbiJ9&platform[accountId]=1`,
+
+	Run: func(cmd *cobra.Command, args []string) {
+		relicString := strings.Join(args, " ")
```
sanderblue, line 21: Does the empty string need the space? 

i.e. can this
```go
strings.Join(args, " ")
```

be this?
```go
strings.Join(args, "")
```

cmd/newrelic/command_decode.go
```diff
@@ -0,0 +1,61 @@
...
+				}
+			}
+
+			decoded, err := base64.StdEncoding.DecodeString(decodedURL)
+			if err != nil {
+				fmt.Println("decode error:", err)
+			}
+
+			decodedURL := string(decoded)
+			fmt.Println("\n**** Decoded Entity ID **** ")
```
sanderblue, line 53: Awesome stuff to get this going. I think we can move forward with this, but we might want to adjust the command signature to be a little more declarative. A declarative signature will help us avoid needing the extra line of output to describe what we're displaying. 

Hypothetical command signature:
```zsh
newrelic decode url https://yadda.yadda.com?pane=eyJuZXJk... --param "pane" --search "entityId"
```

This signature allows the user to define what they want from the URL's query param `pane` as well as what to search for in the encoded string. The `--query-param` flag value can then be passed to the function on [line 23](https://github.com/newrelic/newrelic-cli/pull/653/files#diff-90f45227f18dce8728ec5764cfbef896385ae0d17ac97f9b37ff5d6b64bd2c18R23) and the `--search` value can be passed to [line 34](https://github.com/newrelic/newrelic-cli/pull/653/files#diff-90f45227f18dce8728ec5764cfbef896385ae0d17ac97f9b37ff5d6b64bd2c18R34) and [line 38](https://github.com/newrelic/newrelic-cli/pull/653/files#diff-90f45227f18dce8728ec5764cfbef896385ae0d17ac97f9b37ff5d6b64bd2c18R38) (`gjson.Get(decodedURL, decodeSearch)`) or something to that effect. 

This give the user the ability to decode other URL parts as well if/when things change down the line.

sanderblue (changes requested):
This is looking great! I have one suggestion for the command signature now that I understand the desired output of the command.

zlesnr:
Thank you for the contribution.

PR #656, opened 2021-02-04
feat(decode): NR CLI Translate base64 encoded urls command

GlenOFoghlu:
Created this PR to create Decode CLI operation -> #623


sanderblue (approved):
Awesome! Thanks for the contribution!

PR #662, opened 2021-02-08
chore(install): implement recommended recipe reporting

zlesnr:
Currently, all the recipes up to this point have been of type HOST.  Here we
seek to report on any non-HOST recipes, but only take the install actions when
the type is a HOST.  This work plumbs through the Recommended() methods through
the reporters and ensures that we only attempt to install recipes that are of
type HOST, regardless of what was returned from the recommendations endpoint.

### DIFF
--- a/internal/install/recipe_installer.go
+++ b/internal/install/recipe_installer.go
@@ -241,6 +241,16 @@ func (i *RecipeInstaller) installRecipes(m *types.DiscoveryManifest, recipes []t
 		"recipe_count": len(recipes),
 	}).Debug("installing recipes")
 
+	isHostTarget := func(recipe types.Recipe) bool {
+		for _, target := range recipe.InstallTargets {
+			if target.Type != types.OpenInstallationTargetTypeTypes.HOST {
+				return false
+			}
+		}
+
+		return true
+	}
+
 	for _, r := range recipes {
 		// The infra and logging install have their own install methods.  In the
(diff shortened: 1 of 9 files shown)


internal/install/recipe_installer.go
```diff
@@ -241,6 +241,16 @@ func (i *RecipeInstaller) installRecipes(m *types.DiscoveryManifest, recipes []t
...
+	isHostTarget := func(recipe types.Recipe) bool {
+		for _, target := range recipe.InstallTargets {
+			if target.Type != types.OpenInstallationTargetTypeTypes.HOST {
+				return false
+			}
+		}
+
+		return true
+	}
+
```
jbeveland27, line 253: I think this looks good. My only thought is I wonder if this should be inverted to look for `isApplicationTarget` specifically. This will need to change in the future once we have more than `HOST` and `APPLICATION`, so this is really just a nit at the present moment.

jbeveland27 (approved):

internal/install/recipe_installer.go
jbeveland27: But not necessarily worth holding up this PR over.

zlesnr: I think you're right.  I'll plan to make that change in the other issue where we print the output at the bottom of the install flow, since I think its the logic that you mention, and also solves this case here.

PR #665, opened 2021-02-10
feat(install): allow recipe to display output messages

Julien4218:


zlesnr (changes requested):
Thanks for the PR.  I think we can get rid of the buffer handling above also.

Julien4218:
> Thanks for the PR. I think we can get rid of the buffer handling above also.

Good feedback @zlesnr, I've removed the buffering.

zlesnr (approved):

PR #666, opened 2021-02-10
chore(install): remove the option to skip infra install

zlesnr:
This change means that the infra agent will be installed each time, unless the
user supplies a recipe by name or by URL.


internal/install/recipe_installer.go
```diff
@@ -231,6 +236,25 @@ func (i *RecipeInstaller) Install() error {
 		log.Debugf("Done installing integrations.")
 	}
 
+	isAppTarget := func(recipe types.Recipe) bool {
```
ctrombley, line 239: Maybe worth a convo about this pattern in general, but would this entire func be a bit more readable with the helper in its own scope?

ctrombley (approved):

ctrombley (approved):

ctrombley (approved):

PR #667, opened 2021-02-11
fix(install): make ctrl-c exit reliably

ctrombley:
At any phase during the install command (prompts, execution, validation), if ctrl-c is pressed the command should cease execution without attempting to do anything else.

In the case of third-party libraries, cancelled contexts are handled from within the libraries and allow for executing IO-blocking procedures with a passed context.

All we need to do is handle the error that comes back from those libraries appropriately.  In the case of the OHI execution loop, we opt to try the next recipe if the current one fails - this leads to unexpected behavior when ctrl-c is the cause of the error.

This PR makes it so that we will try the next recipe, but only if the error that comes back from the various 3rd party libraries was not generated by a context cancellation.

### DIFF
--- a/internal/install/execution/go_task_recipe_executor.go
+++ b/internal/install/execution/go_task_recipe_executor.go
@@ -7,6 +7,7 @@ import (
 	"io/ioutil"
 	"os"
 	"strconv"
+	"strings"
 
 	"github.com/go-task/task/v3"
 	taskargs "github.com/go-task/task/v3/args"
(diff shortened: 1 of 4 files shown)


internal/install/execution/go_task_recipe_executor.go
```diff
@@ -204,6 +211,10 @@ func varsFromInput(inputVars []recipes.VariableConfig, assumeYes bool) (types.Re
 
 			envValue, err = varFromPrompt(envConfig)
 			if err != nil {
+				if err == promptui.ErrInterrupt {
```
zlesnr, line 214: 👍🏼

internal/utils/utils.go
```diff
@@ -25,7 +25,7 @@ func getSignalContext() context.Context {
 	signal.Notify(ch, syscall.SIGINT, syscall.SIGTERM)
 	go func() {
 		sig := <-ch
-		log.Warnf("signal received: %s", sig)
+		log.Debugf("signal received: %s", sig)
```
zlesnr, line 28: 👍🏼

zlesnr (approved):
Looks good, nice work.

PR #671, opened 2021-02-11
fix(install): Ensure error bubbles up when executing only 1 recipe and failing

Julien4218:


zlesnr (commented):
This makes sense, though I think this might be better served by updating the terminal_reporter, since that is where the user gets told at the console that there was an error.

ctrombley:
@zlesnr Regardless of what message we display to the user, I think we need to return an error here to exit with a non-zero code.

ctrombley (approved):

PR #673, opened 2021-02-12
chore(install): add a preinstall message

ctrombley:


internal/install/recipe_installer.go
```diff
@@ -502,6 +502,10 @@ func (i *RecipeInstaller) executeAndValidate(m *types.DiscoveryManifest, r *type
 }
 
 func (i *RecipeInstaller) executeAndValidateWithProgress(m *types.DiscoveryManifest, r *types.Recipe) (string, error) {
+	if r.PreInstall.Info != "" {
+		fmt.Println(r.PreInstall.Info)
```
zlesnr, line 506: We might want to pad this with two spaces before the message.  Same for the `Println` call be low.

zlesnr (dismissed):

internal/install/recipe_installer.go
jbeveland27: @zlesnr Since most of these will be multi-line strings, should we just do that padding in the source recipe?

jbeveland27: Basically we need to achieve things like this:

image

jbeveland27: So wondering if we let the CLI provide indentation from the left, or if we just craft all our strings with two spaces of indentation.

zlesnr: That's a good point.  We could take it as is and see what the feedback is too.

jbeveland27 (approved):
LGTM 🚀

PR #675, opened 2021-02-12
chore(install): make insights insert key available to recipes

ctrombley:


Julien4218 (commented):
@ctrombley maybe we should remove insight insert api key for now?

zlesnr (dismissed):

zlesnr:
Can you elaborate @Julien4218?  Maybe I'm missing the context.

Julien4218:
> Can you elaborate @Julien4218? Maybe I'm missing the context.

Yes, this was a separate thread in slack.
So it looks like the insight insert api key can only be generated through the UI. So we may not have an actual value at the moment. Also there is a suspicion there might a migration to the ohi currently requiring this key to maybe use the more widely use personal api key in the future. So perhaps this particular key might never be needed.
Given rishi mentions to stop/hold on that specific ohi, I dont think we need the insight key at this time.

PR #699, opened 2021-02-20
chore(deps): bump go to v1.16

chenrui333:
👋  go 1.16 use go-modules build by default, since this is CLI app, there is no reason to do multiple go versions support. Hence, I have dropped the old go versions.


zlesnr:
Hello, thank you for the PR.  That sounds right to me also.  It was probably a cary-over from another project, but I see your point.  I'll upgrade my local system and run the tests.

zlesnr:
It doesn't look like brew has 1.16 yet so I didn't test locally.  I think we want to hold this open for a week or so for reasons I can't share yet, but we're on board with this change.

chenrui333:
revisit this PR today

chenrui333:
relates to [the new homebrew bump](https://github.com/Homebrew/homebrew-core/pull/79542)

ctrombley (approved):

ctrombley:
Thanks @chenrui333 , we will continuing this effort in #972 .

Issue #702, opened 2021-02-22
Unneeded text in README.md for Chocolatey


kidk:
### Description
The [readme](https://github.com/newrelic/newrelic-cli/blob/master/README.md) states the following for Chocolatey, but it doesn't match anymore with the command below.

```
Currently the `newrelic-cli` is in review with in the Chocolatey system, and
should be published in the next few days.  That means that you must include the
version spec, as follows.
```

I wanted to double check if it's ok to remove? Happy to create the pull request.

zlesnr:
Thank you for the note.  That does look like it needs an update, PRs welcome.

kidk:
Pull request created 👆 Thanks

PR #703, opened 2021-02-22
Fix: removed outdated warning for Chocolatey

kidk:
The warning isn't needed anymore. Ref: #702


ctrombley (approved):
Thanks for your contribution @kidk!

PR #705, opened 2021-02-22
fix(install): Change fatal error message when failing to install

Julien4218:


sanderblue (approved):

internal/install/command.go
```diff
@@ -56,7 +55,7 @@ var Command = &cobra.Command{
 
 			// Run the install.
 			if err := i.Install(); err != nil {
-				log.Fatalf("Could not install New Relic: %s, check the install log for details: %s", err, config.DefaultLogFile)
+				log.Fatalf("We encountered an error with the installation, if this problem persists please visit the documentation and support page for additional help here: https://one.nr/06vjAeZLKjP")
```
zlesnr, line 58: Consider changing "with the installation" to "during installation".  Also here, the "if this" should be a new sentence.

zlesnr (dismissed):

internal/install/command.go
Julien4218: I've copy/pasted what Rishi requested

zlesnr: As written, it looks like two sentences.  I approved the PR.

Julien4218: Yeah, I agree. I've updated.

PR #714, opened 2021-02-24
Chore/report all entities

ctrombley:
This PR resolves both VIRTUOSO-380 (introduce a spinner during validation) and VIRTUOSO-381 (report entity-scoped data for all known entities during guided install).  The changes are staged across two separate commits for ease of review.

### Testing:
- For VIRTUOSO-380, `./bin/darwin/newrelic installTest` can be used to validate a mock scenario.  Next, running the installer on an actual VM should yield the same UX updates with a spinner included.
- For VIRTUOSO-381, a successful redis OHI install should no longer report redis as an available integration on the entity overview page after a successful installation.

### DIFF
--- a/internal/install/ux/spinner.go
+++ b/internal/install/ux/spinner.go
@@ -6,14 +6,13 @@ import (
 
 	spinnerLib "github.com/briandowns/spinner"
 	log "github.com/sirupsen/logrus"
-
-	"github.com/newrelic/newrelic-cli/internal/install/types"
 )
 
 const (
-	interval  = 100 * time.Millisecond
-	checkmark = "\u2705"
-	crossmark = "\u274C"
+	interval    = 100 * time.Millisecond
+	checkmark   = "\u2705"
+	crossmark   = "\u274C"
+	indentation = "    "
 )
 
 var (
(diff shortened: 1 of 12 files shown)


internal/install/ux/spinner.go
```diff
@@ -52,10 +50,12 @@ func (s *Spinner) Stop() {
...
+func (s *Spinner) Fail(msg string) {
+	s.FinalMSG = indentation + crossmark
+	s.Suffix = s.Suffix + "failed."
 }
 
-func (s *Spinner) Success(recipe types.Recipe) {
-	s.FinalMSG = checkmark
+func (s *Spinner) Success(msg string) {
+	s.FinalMSG = indentation + checkmark
+	s.Suffix = s.Suffix + "success."
```
sanderblue, line 60: Will this cause `s.Suffix` to miss a space between it and 'success'?

zlesnr (approved):

internal/install/ux/spinner.go
ctrombley: The intended text is "...success.", so no space is required.

PR #726, opened 2021-03-01
fix(install): allow skipDiscovery and skipLoggingInstall flags to work together

ctrombley:


Julien4218:
@ctrombley I'm going to test this one a few use cases, please hold on merging.

Julien4218 (approved):
This works well. I've tested various installs guided with and without log, with and without integrations, and stitch path. It all works.

PR #731, opened 2021-03-02
chore(install): add infra installation to the targeted install

Julien4218:


ctrombley (approved):

zlesnr:
Thanks for putting up the PR @Julien4218.  Are we sure this is desirable?  This seems like it might surprise the user.

Julien4218:
Yes @zlesnr there are definitely some pros/cons either way we go. We've detailed some of those in the ticket https://newrelic.atlassian.net/browse/VIRTUOSO-400
We can either not install the infra agent, and the stitch-path install will fail (current state).
We can install the infra agent if we detect no infra agent is install when installing an OHI/stitch-path. If we chose that, then potentially the OHI we're installing is incompatible with the version of the infra agent installed (too old possibly).
Finally, we can always install the latest version of the infra agent when installing an OHI/stich-path. We've chosen that path. The possible downside, if there are already some other OHI installed that relied on potentially an older version of the infra agent (which was supposedly there before we install the latest) then that older installed OHI may no longer be working with the most recent infra agent version. This however should be fixed by re-running an install for that older OHI.

PR #735, opened 2021-03-04
fix: Replace Detected observability gaps by Data Gaps

idirouhab:
The UI says Data gap and not Detected observability gaps

image


zlesnr (approved):
Thank you for the PR.

PR #736, opened 2021-03-04
chore(install): capture canceled status

sanderblue:


internal/install/command.go
```diff
@@ -55,6 +57,11 @@ var Command = &cobra.Command{
 
 			// Run the install.
 			if err := i.Install(); err != nil {
+				if err == types.ErrInterrupt {
+					fmt.Println("DS:DNFLKSDJFNLKJSDNFLJSKDNFKLJSDNFKLJNFLJSKDNFKJLn")
```
ctrombley, line 61: 🤣

internal/install/execution/install_status.go
```diff
@@ -187,6 +199,16 @@ func (s *InstallStatus) hasFailed() bool {
 	return false
 }
 
+func (s *InstallStatus) isCanceled() bool {
```
ctrombley, line 202: Refactor opportunity: this could be refactored to a general purpose method to test any state, and `isCanceled` and `isCompleted` could then be based on it to DRY things up a little.

internal/install/scenario_builder.go
```diff
@@ -331,3 +372,64 @@ install:
...
+    run:
+      cmds:
+        - |
+          echo ""
+          echo "ctl+c right MEOW!!!!"
+          echo ""
+
+          sleep 10
+
+postInstall:
```
ctrombley, line 425: this postInstall is unnecessary for testing this scenario.

internal/install/command.go
Julien4218: what?

internal/install/execution/install_status.go
Julien4218: I like that idea

Julien4218:
This looks good, let me know when it's ready to test, I'll try it out.

Out of curiosity, what would it take to drive a Cancelled status for a specific error code value? I'm thinking we might find use case soon when a recipe might want to trigger a cancel without the user doing CTRL+C, doing so with a specific exit code value could be a way to achieve that.

internal/install/command.go
ctrombley: I don't think `fmt.Println("DS:DNFLKSDJFNLKJSDNFLJSKDNFKLJSDNFKLJNFLJSKDNFKJLn")` is needed here 😉

zlesnr:
Implementing special errors codes based on exit code would be simple after this PR.  Here we lay the foundation for just such a case.

zlesnr (approved):

Julien4218 (approved):
Tested on a ubuntu with infra/log + 2 OHI.
Run first with current released CLI then re-run with this version to see the difference in the nerdstorage document, looking good.

Julien4218:
[ctrlc.txt](https://github.com/newrelic/newrelic-cli/files/6104893/ctrlc.txt)

Julien4218:
1 important thing to note, the recommendation disappears from the explorer/host recommended nerdlet view.
For example, I could see 'Apache' was recommended before, but I ctrl+c to stop it, now it doesn't show in the UI as a recommended install.

PR #746, opened 2021-03-11
chore(execution): update terminal report GUID when targeted install

zlesnr:


sanderblue (approved):
🐑 it

Julien4218 (approved):
This works for the CLI entity URL displayed.
This does not work for the UI "See your Data" button (goes to host entity in Explorer view).

PR #747, opened 2021-03-12
fix(install): avoid error return when OHI recipe fails

zlesnr:


internal/install/recipe_installer_guided.go
```diff
@@ -16,6 +16,10 @@ const (
 	loggingRecipeName    = "logs-integration"
 )
 
+// guidedInstall walks the user through and installation, prompting for input
+// when needed.  An error is returned only when the infra or logging recipes
+// have an error.  Of an OHI recipe fails, we warn the user.  This allows the
```
ctrombley, line 21: nit: s/Of/If

internal/install/recipe_installer_guided.go
```diff
@@ -107,7 +111,7 @@ func (i *RecipeInstaller) guidedInstall(ctx context.Context, m *types.DiscoveryM
 	if i.ShouldInstallIntegrations() {
 		log.Debugf("Installing integrations")
 		if err = i.installRecipes(ctx, m, selectedIntegrations); err != nil {
```
ctrombley, line 113: should we just change this line to `_ = i.installRecipes(ctx, m, selectedIntegrations)`?

zlesnr: That's not a bad idea.  I'm now seeing the debug log that comes after this, so, should we log something about receiving errors here?

zlesnr: I've updated the logic here to keep the ErrInterrupt handled correctly.

ctrombley (approved):
:shipit:

PR #751, opened 2021-03-15
fix(install): ensure grep sed awk cat tee touch are installed

Julien4218:


zlesnr:
This feels a little disconnected, since the new requirements aren't used in the specific script, though they are used in the ecosystem.  If we left them out, we could at least get the CLI installed, which has other uses than the `install` command.  Then we could work something into the `install` command to warn for the specific cases.  What do you think?

Julien4218:
> This feels a little disconnected, since the new requirements aren't used in the specific script, though they are used in the ecosystem. If we left them out, we could at least get the CLI installed, which has other uses than the `install` command. Then we could work something into the `install` command to warn for the specific cases. What do you think?

Correct, I was expecting this feedback. It's tricky. I liked this part to try to front the requirement before we even install the CLI, but could see how those requirements are specific to the install which is a more focus usage of the CLI. I don't mind not implementing this work for now, and find a better way to implement it in the install flow in the future, perhaps once we have data to illustrate those dependencies are causing issues.

PR #752, opened 2021-03-15
fix(install): add a better error for 404s when loading recipe files

ctrombley:


internal/install/recipes/recipe_file.go
```diff
@@ -69,6 +70,11 @@ func (f *RecipeFileFetcherImpl) FetchRecipeFile(recipeURL *url.URL) (*RecipeFile
 	if err != nil {
 		return nil, err
 	}
+
+	if response.StatusCode != 200 {
```
Julien4218, line 74: We should check the whole 2xx range?

ctrombley: 👍

Julien4218 (dismissed):

michaelkro (approved):
🚢

zlesnr (approved):

zlesnr (approved):

PR #755, opened 2021-03-16
chore(execution): include CLI version in nerdstorage payload

zlesnr:


cmd/newrelic/main.go
```diff
@@ -50,6 +52,8 @@ func init() {
 	Command.AddCommand(apiaccess.Command)
 
 	CheckPrereleaseMode(Command)
+
+	os.Setenv("CLI_VERSION", version)
```
Julien4218, line 56: should we use the consistent NEW_RELIC_ prefix syntax?

zlesnr: Good call, I'll update.

Julien4218 (approved):

PR #756, opened 2021-03-16
chore(execution): include validation duration in RecipeStatus

zlesnr:


internal/install/execution/install_status.go
```diff
@@ -28,11 +28,12 @@ type InstallStatus struct {
...
-	Error       StatusError      `json:"error"`
-	Name        string           `json:"name"`
-	Status      RecipeStatusType `json:"status"`
-	EntityGUID  string           `json:"entityGuid,omitempty"`
+	DisplayName        string           `json:"displayName"`
+	Error              StatusError      `json:"error"`
+	Name               string           `json:"name"`
+	Status             RecipeStatusType `json:"status"`
+	EntityGUID         string           `json:"entityGuid,omitempty"`
+	ValidationDuration int64            `json:"validationDuration,omitempty"`
```
Julien4218, line 36: I would really encourage to put the unit as a suffix on that field, otherwise users will keep wondering what the unit is, it's really unclear, could be seconds, milli, micro, ticks ... whatever

zlesnr: Would it be sufficient to add documentation to this field?

Julien4218: I would think not. The best documentation is the implementation in my mind: 1) it never gets outdated, 2) no need to search for an eventual documentation dissociated from where I use or read the field (crafting nrql query for example)

zlesnr: Updated to match the documentation.

Julien4218 (approved):

PR #768, opened 2021-03-22
chore(install): add debug for logs variable and only write DISCOVERED_LOG_FILES if defined

Julien4218:


Julien4218:
While fixing and troubleshooting the logs recipes, it would be helpful to have the extra debug info for the logs variable.
Also, as a suggestion, this removes the DISCOVERED_LOG_FILES variable for the recipe that don't define any log matches (most recipes).

zlesnr (approved):

PR #771, opened 2021-03-23
chore(install): update recipe schema with successLink field

ctrombley:
Updating `successLink` -> `successLinkConfig` and updating docs per UAPI feedback.

cc @michaelkro

### DIFF
--- a/internal/install/execution/install_status.go
+++ b/internal/install/execution/install_status.go
@@ -24,7 +24,7 @@ type InstallStatus struct {
 	DocumentID        string
 	targetedInstall   bool
 	statusSubscriber  []StatusSubscriber
-	successLink       types.SuccessLink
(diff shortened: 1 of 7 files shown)


zlesnr (approved):

jbeveland27 (approved):

jbeveland27:
@zlesnr Looking for confirmation here; I think we need to update this [query](https://github.com/newrelic/newrelic-cli/blob/master/internal/install/recipes/service_recipe_fetcher.go#L282) to actually get `successLinkConfig` from NerdGraph. Should the sequencing be that we merge this PR first, and then update the query after it's been stitched into NerdGraph? (otherwise we risk the NerdGraph call just flat out bailing since we're asking for an invalid field until it's stitched)

PR #772, opened 2021-03-23
fix: avoid nil pointer when fetching a license key

zlesnr:


cmd/newrelic/command.go
```diff
@@ -151,9 +151,12 @@ func fetchLicenseKey(client *newrelic.NewRelic, accountID int) (string, error) {
...
 	actor := queryResp.Actor.(map[string]interface{})
 	account := actor["account"].(map[string]interface{})
-	licenseKey := account["licenseKey"].(string)
 
-	return licenseKey, nil
+	if licenseKey, ok := account["licenseKey"]; ok {
+		return licenseKey.(string), nil
+	}
+
+	return "", fmt.Errorf("license key unavailable")
```
Julien4218, line 159: Should we have a more friendly user message?

zlesnr: Yes probably.  The caller throws away the error currently, so I'm still trying to reproduce the specific case where this comes up.  Without knowing the context as to why the license key was not returned, I'm not quite sure what to tell the user, or if we should fail the run, or if we should retry.

Julien4218: Right, lots of good thinking to do.
I think we should assume we have some sort of sev events, which is likely to be resolved soon. We can check with rishi and the UX folks. I think a message along the line of "oops, we're having some technical difficulties, please retry later" and even perhaps giving an approximate timeline (1 hour should be confortable).
Alternatively and/or in addition, we could surface our manual install steps, which is what we currently surface when a recipe fails with this URL https://docs.newrelic.com/docs/using-new-relic/cross-product-functions/troubleshooting/not-seeing-data/

zlesnr: I've updated the message here, and implemented a retry as well as log the error when it is returned.  Maybe should sleep between retries.

sanderblue (approved):

Julien4218 (approved):

PR #776, opened 2021-03-24
feat(install): additional queryable fields for InstallStatus event

sanderblue:


internal/install/execution/install_status.go
```diff
@@ -381,3 +375,65 @@ func (s *InstallStatus) getStatus(r types.Recipe) *RecipeStatus {
 
 	return nil
 }
+
+// This function handles updating the final recipe statuses and top-level installation status.
+// Canceling (e.g. ctl+c) will cause unresolved recipes to be marked as canceled.
+// Exiting early (i.e. an error occurred) will cause unresolved recipes to be marked as failed.
+func (s *InstallStatus) updateFinalInstallationStatuses(installCanceled bool) {
```
sanderblue, line 382: Self review: Needing to pass a boolean (or some indicator/value) to differentiate between canceled/complete to this method presented some code smell. I tried to DRY up some of the repeated code in the `for` loop. Doing so left me thinking we might want to consider consolidating `InstallStatus.InstallCanceled()` and `InstallStatus.InstallComplete()` into a single method that captures the idea of "install exited" and then handle setting "failed" vs "canceled" in there, but even that seems like we can make setting the installation and recipe states in a cleaner way in a future PR.

internal/install/execution/install_status.go
```diff
@@ -381,3 +374,59 @@ func (s *InstallStatus) getStatus(r types.Recipe) *RecipeStatus {
...
```
sanderblue, line 381: Self review: Needing to pass a boolean (or some indicator/value) to differentiate between canceled/complete to this method presented some code smell. I tried to DRY up some of the repeated code in the for loop. Doing so left me thinking we might want to consider consolidating InstallStatus.InstallCanceled() and InstallStatus.InstallComplete() into a single method that captures the idea of "install exited" and then handle setting "failed" vs "canceled" in there, but even that seems like we can make setting the installation and recipe states in a cleaner way in a future PR.

internal/install/execution/install_status.go
```diff
@@ -13,18 +14,26 @@ import (
...
+	Timestamp           int64                   `json:"timestamp"`
+	CLIVersion          string                  `json:"cliVersion"`
+	HasInstalledRecipes bool                    `json:"hasInstalledRecipes"`
+	HasCanceledRecipes  bool                    `json:"hasCanceledRecipes"`
+	HasSkippedRecipes   bool                    `json:"hasSkippedRecipes"`
+	HasFailedRecipes    bool                    `json:"hasFailedRecipes"`
+	RecipesSkipped      []*RecipeStatus         `json:"recipesSkipped"`
+	RecipesCanceled     []*RecipeStatus         `json:"recipesCanceled"`
+	RecipesFailed       []*RecipeStatus         `json:"recipesFailed"`
+	RecipesInstalled    []*RecipeStatus         `json:"recipesInstalled"`
```
sanderblue, line 32: The `RecipesSkipped`, `RecipesCanceled`, `RecipesFailed`, and `RecipesInstalled` arrays don't necessarily need to be here, but figured I'd throw the idea out there in case this data structure might be helpful. If not, they can safely be removed. Thoughts?

internal/install/execution/install_status.go
zlesnr: I'm thinking this could be a good use of copying the pattern with `RecipeStatusType`, where we could have a `FinalStatusType` that this method could receive.  Then the caller could just pass the desired final and delete the two methods like you mentioned.

zlesnr (approved):

PR #778, opened 2021-03-29
Fix/fetch license key

Julien4218:


internal/install/execution/go_task_recipe_executor.go
```diff
@@ -137,15 +137,15 @@ func (re *GoTaskRecipeExecutor) Execute(ctx context.Context, m types.DiscoveryMa
 	return nil
 }
 
-func varsFromProfile() (types.RecipeVars, error) {
+func varsFromProfile(licenseKey string) (types.RecipeVars, error) {
 	defaultProfile := credentials.DefaultProfile()
 	if defaultProfile.LicenseKey == "" {
```
ctrombley, line 142: This check probably no longer makes sense if we want to move ahead with this change.

internal/install/recipe_installer.go
```diff
@@ -235,7 +237,12 @@ func (i *RecipeInstaller) executeAndValidateWithProgress(ctx context.Context, m
 		fmt.Println(r.PreInstallMessage())
 	}
 
-	vars, err := i.recipeExecutor.Prepare(ctx, *m, *r, i.AssumeYes)
+	licenseKey, err := i.licenseKeyFetcher.FetchLicenseKey(ctx)
+	if err != nil {
+		return "", err
+	}
+
+	vars, err := i.recipeExecutor.Prepare(ctx, *m, *r, i.AssumeYes, licenseKey)
```
ctrombley, line 245: This `Prepare` method seems to be increasingly meaningless as a contract especially as we add more arguments to it.  Would it be better to add the licenseKeyFetcher as a dependency on the recipeExecutor instead, and do the fetching from within?

internal/install/license_key_fetcher.go
```diff
@@ -0,0 +1,65 @@
...
+
+func NewServiceLicenseKeyFetcher(client recipes.NerdGraphClient) LicenseKeyFetcher {
+	f := ServiceLicenseKeyFetcher{
+		client: client,
+	}
+
+	return &f
+}
+
+func (f *ServiceLicenseKeyFetcher) FetchLicenseKey(ctx context.Context) (string, error) {
```
ctrombley, line 30: This code is mostly a duplicate of this init code that is run the first time a CLI command is executed after installation:
https://github.com/newrelic/newrelic-cli/blob/master/cmd/newrelic/command.go#L142-L167 

Could be an opportunity to DRY up the code.

ctrombley:
This looks like it will fix the problem, but is a bit awkward in the code since we have a license key defined in the profile and we are choosing to ignore it.

Another approach might be to advise the user if their license key doesn't match their API key, but that would also be putting more responsibility on the user to fix that problem themselves.

internal/install/execution/go_task_recipe_executor.go
Julien4218: Good catch, I've updated the assertion to validate with the provided license key

Julien4218:
> This looks like it will fix the problem, but is a bit awkward in the code since we have a license key defined in the profile and we are choosing to ignore it.
> 
> Another approach might be to advise the user if their license key doesn't match their API key, but that would also be putting more responsibility on the user to fix that problem themselves.

I agree. I think this should fix the issue we're aware of with license key, and also open the door to re-evaluate in subsequent effort how to manage the CLI credentials.
Let me know if you have any objections moving this change forward, I'd like to fix the issues we're aware of as soon as possible.

ctrombley (approved):

PR #790, opened 2021-04-01
fix(targeted-install): include successLinkConfig in Recipe

jbeveland27:


gabeobrien (approved):
Looks good and works!

PR #792, opened 2021-04-07
fix(install): discovered log files needs to be a string to work within our recipe

sanderblue:
Related: [newrelic/open-install-library #293](https://github.com/newrelic/open-install-library/pull/293)

### DIFF
--- a/internal/install/recipe_installer_guided.go
+++ b/internal/install/recipe_installer_guided.go
@@ -154,16 +154,26 @@ func (i *RecipeInstaller) installLogging(ctx context.Context, m *types.Discovery
 		"matches": acceptedLogMatches,
 	}).Debug("matches accepted")
 
+	// The struct to approximate the logging configuration file of the Infra Agent. (deprecated)
+	type loggingConfig struct {
+		Logs []types.LogMatch `yaml:"logs"`
+	}
+	// Deprecated. Will be removed in a future release and replaced by NR_DISCOVERED_LOG_FILES.
+	// We need to keep this var for backwards compatibility until recipes have been updated with the new var.
+	r.AddVar("DISCOVERED_LOG_FILES", loggingConfig{Logs: acceptedLogMatches})
+
+	// Build a comma-separated list of discovered log file paths
 	discoveredLogFiles := []string{}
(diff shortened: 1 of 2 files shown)


ctrombley (approved):

zlesnr (dismissed):

internal/install/recipe_installer_guided.go
```diff
@@ -153,12 +154,17 @@ func (i *RecipeInstaller) installLogging(ctx context.Context, m *types.Discovery
...
-	// The struct to approximate the logging configuration file of the Infra Agent.
-	type loggingConfig struct {
-		Logs []types.LogMatch `yaml:"logs"`
+	discoveredLogFiles := []string{}
+	for _, logMatch := range acceptedLogMatches {
+		discoveredLogFiles = append(discoveredLogFiles, logMatch.File)
 	}
 
-	r.AddVar("DISCOVERED_LOG_FILES", loggingConfig{Logs: acceptedLogMatches})
+	discoveredLogFilesString := strings.Join(discoveredLogFiles, ",")
```
Julien4218, line 162: It's tricky to get this deployed with the library changes.
Could we add a new data element maybe named "NR_DISCOVERED_LOG_FILES" in addition of the existing "DISCOVERED_LOG_FILES", and then in a subsequent release remove the old element "DISCOVERED_LOG_FILES" (after all recipes have been updated to use the new data element).

Julien4218 (commented):
@sanderblue I like the change, we'll have a better time with the recipe dealing with a CSV list of filepath than a yaml structure.

However, can we phase the CLI change into 2 releases, adding a new data element while preserving the existing "DISCOVERED_LOG_FILES", and then doing another release later to remove the old data element?

internal/install/recipe_installer_guided.go
sanderblue: Oh right, thanks for catching that.

sanderblue: Updated 😎

zlesnr (approved):

Julien4218 (approved):
Looks good

PR #793, opened 2021-04-07
Chore/assert platform supported

Julien4218:
We've validated the correct error message when running a darwin machine.
We also see an event INSTALL_GuidedInstall_completeInstall in Tessen, with the os column of darwin, however we don't see any error message. There might be something else to do in the UI to wire through the error message

### DIFF
--- a/internal/install/execution/install_status.go
+++ b/internal/install/execution/install_status.go
@@ -175,19 +175,8 @@ func (s *InstallStatus) RecipeSkipped(event RecipeStatusEvent) {
 	}
 }
 
-func (s *InstallStatus) InstallError(err error) {
-	statusError := StatusError{
-		Message: "unknown error",
-	}
-	if err != nil {
-		statusError.Message = err.Error()
-	}
-	s.Error = statusError
-	s.InstallComplete()
-}
-
-func (s *InstallStatus) InstallComplete() {
-	s.completed()
+func (s *InstallStatus) InstallComplete(err error) {
+	s.completed(err)
 
 	for _, r := range s.statusSubscriber {
 		if err := r.InstallComplete(s); err != nil {
@@ -355,9 +344,16 @@ func (s *InstallStatus) withRecipeEvent(e RecipeStatusEvent, rs RecipeStatusType
 	s.Timestamp = utils.GetTimestamp()
 }
 
-func (s *InstallStatus) completed() {
+func (s *InstallStatus) completed(err error) {
 	s.Complete = true
 	s.Timestamp = utils.GetTimestamp()
+
+	if err != nil {
+		statusError := StatusError{
+			Message: err.Error(),
+		}
+		s.Error = statusError
+	}
 
 	log.WithFields(log.Fields{
 		"timestamp": s.Timestamp,
--- a/internal/install/execution/install_status.go
+++ b/internal/install/execution/install_status.go
@@ -175,6 +175,17 @@ func (s *InstallStatus) RecipeSkipped(event RecipeStatusEvent) {
 	}
 }
 
+func (s *InstallStatus) InstallError(err error) {
+	statusError := StatusError{
+		Message: "unknown error",
+	}
+	if err != nil {
+		statusError.Message = err.Error()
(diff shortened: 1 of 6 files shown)


ctrombley (approved):
LGTM

Julien4218:
@ctrombley / @zlesnr 
I made an additional change to try to surface any error message to the nerdstorage, so it can be surfaced in the tessen event (the UI will if the data is there).
Could you check if what I've done is correct to get an error message when the install fails for any reason (other than ctrl+c)?

internal/install/execution/install_status.go
```diff
@@ -175,6 +175,17 @@ func (s *InstallStatus) RecipeSkipped(event RecipeStatusEvent) {
 	}
 }
 
+func (s *InstallStatus) InstallError(err error) {
```
zlesnr, line 178: It looks to me like if `err` is nil, then the status will still contain `unknown error` here.  I'm wondering if we could smooth out the error handling here.  Perhaps `InstallComplete()` could take the error here rather than a separate method for it.

internal/install/recipe_installer.go
```diff
@@ -138,6 +144,17 @@ func (i *RecipeInstaller) discoverAndRun(ctx context.Context) error {
 	return i.guidedInstall(ctx, m)
 }
 
+func (i *RecipeInstaller) assertDiscoveryValid(ctx context.Context, m *types.DiscoveryManifest) error {
+	if m.OS == "" {
+		return fmt.Errorf("failed to identify a valid operating system")
```
zlesnr, line 149: We've started putting errors in the `internal/install/types` package.  What do you think about moving these over there and then just returning the particular error here?

internal/install/recipe_installer.go
```diff
@@ -138,6 +144,17 @@ func (i *RecipeInstaller) discoverAndRun(ctx context.Context) error {
...
+	}
+	if !(strings.ToLower(m.OS) == "linux" || strings.ToLower(m.OS) == "windows") {
+		return fmt.Errorf("operating system %s is not supported", m.OS)
+	}
+	log.Debugf("Done asserting valid operating system for %s", m.OS)
+	return nil
+}
```
jbeveland27, line 156: I think we still need a check in here for when `platform` is empty

internal/install/execution/install_status.go
Julien4218: @zlesnr 
I'm not quite understand go enough here (channel wiring), can you tell me what would cause the caller to get into this flow with a nil error?
I know the tests did that, hence the nil check, is there a valid usage when installing where we would error without an actual error object?

internal/install/recipe_installer.go
Julien4218: I'm gonna refrain from doing this at this time

internal/install/recipe_installer.go
Julien4218: @jbeveland27 I'm not sure about an empty check on platform. I could see recipes just defining OS without platform.
We can always add more rules later when we're sure. For now, I'd like to get this enhancement out to protect the cases we know are not supported (darwin).

Julien4218:
Tested on a macos machine, and looking at the event data in Tessen I can now see an error_message "operating system darwin is not supported" for the event "INSTALL_GuidedInstall_completeInstall"

internal/install/execution/install_status.go
ctrombley: `InstallComplete` method is meant to handle the success case as well as the error case.  In the success case, `err` will be nil.  I would get rid of this `InstallError` method, and instead do this check in `InstallComplete`:

```
if err != nil, set the nerdstorage error
```

ctrombley (changes requested):

internal/install/execution/install_status.go
Julien4218: Ok will do, thanks @ctrombley

zlesnr (dismissed):

ctrombley (approved):

PR #795, opened 2021-04-12
chore: drop promptui dependency, replace with survey

zlesnr:


sanderblue (approved):
🐑 it

internal/install/execution/go_task_recipe_executor.go
```diff
@@ -233,26 +234,31 @@ func varFromPrompt(envConfig recipes.VariableConfig) (string, error) {
...
-	}
+	value := ""
+	var prompt survey.Prompt
 
 	if envConfig.Secret {
-		prompt.HideEntered = true
-		prompt.Mask = '*'
+		prompt = &survey.Password{
+			Message: msg,
+		}
```
jbeveland27, line 243: Does this automagically mask the input with `*****`?

zlesnr: It does.  You can see an example here: https://github.com/AlecAivazis/survey/#password

zlesnr: Feel free to check the branch out and give it a test though.

jbeveland27: Ah cool 🙌🏼

internal/install/execution/go_task_recipe_executor.go
```diff
@@ -233,26 +234,31 @@ func varFromPrompt(envConfig recipes.VariableConfig) (string, error) {
...
+		defaultValue := ""
+
+		if envConfig.Default != "" {
+			defaultValue = envConfig.Default
+		}
+
+		prompt = &survey.Input{
+			Message: msg,
+			Default: defaultValue,
+		}
```
jbeveland27, line 254: Would be interesting to consider using the [suggest options](https://github.com/AlecAivazis/survey#suggestion-options) feature vs what we do now with having the default as a part of the [prompt message](https://github.com/newrelic/open-install-library/blob/main/recipes/newrelic/infrastructure/ohi/apache/debian.yml#L40)

jbeveland27: That'd involve a change to the schema, so wouldn't happen right away. But might offer a better UX.

zlesnr: Yeah, that sounds nice.  Perhaps in another pr?

jbeveland27: Yup, not needed for this one, but possible future enhancement.

jbeveland27 (approved):
LGTM 🚢

PR #799, opened 2021-04-12
chore(install): add validation rules for older versions of windows

Julien4218:


Julien4218:
https://newrelic.atlassian.net/browse/VIRTUOSO-640

internal/install/discovery/os_windows_validator.go
```diff
@@ -0,0 +1,62 @@
...
+			if ierr == nil {
+				return ensureMinimumVersion(major, minor)
+			}
+		}
+	}
+
+	return WindowsNoVersionMessage
+}
+
+func ensureMinimumVersion(major int, minor int) string {
```
zlesnr, line 52: Instead of returning a string here, returning an error as we do in the `types` package I think would be more appropriate.  Then you could return `nil` for the case that it doesn't match.

internal/install/discovery/os_windows_validator.go
```diff
@@ -0,0 +1,62 @@
...
+	WindowsNoVersionMessage         = "Failed to identified a valid version of Windows"
+)
+
+func NewOsWindowsValidator() *OsWindowsValidator {
+	validator := OsWindowsValidator{}
+
+	return &validator
+}
+
+func (v *OsWindowsValidator) Execute(m *types.DiscoveryManifest) string {
```
zlesnr, line 23: Again here, it seems that the return should be an error.

internal/install/recipe_installer_test.go
```diff
@@ -132,75 +134,12 @@ func TestInstall_FailsOnInvalidOs(t *testing.T) {
...
 	}
 
-	i := RecipeInstaller{ic, discover, l, f, e, v, ff, status, p, pi, lkf}
+	i := RecipeInstaller{ic, discover, l, mv, f, e, v, ff, status, p, pi, lkf}
 
 	err := i.Install()
 	require.Error(t, err)
 }
 
-func TestInstall_FailsOnNoOs(t *testing.T) {
```
zlesnr, line 141: Do we not want to fail if the manifest contains an empty OS anymore?

internal/install/discovery/manifest_validator.go
```diff
@@ -0,0 +1,61 @@
...
+		validators: []Validator{},
+	}
+
+	mv.validators = append(mv.validators, NewOsValidator())
+	mv.validators = append(mv.validators, NewOsWindowsValidator())
+
+	return &mv
+}
+
+func (mv *ManifestValidator) Execute(m *types.DiscoveryManifest) string {
```
jbeveland27, line 34: I think @zlesnr said the same elsewhere, but feels like these errors should be propagated up and returned as error vs string.

internal/install/discovery/os_validator.go
```diff
@@ -0,0 +1,33 @@
...
+	"fmt"
+	"strings"
+
+	"github.com/newrelic/newrelic-cli/internal/install/types"
+)
+
+type OsValidator struct{}
+
+var (
+	NoOperatingSystemDetected         = "failed to identify a valid operating system"
```
sanderblue, line 13: Do we need to export these variables? i.e. if not, they can start with a lowercase letter.

Julien4218: Good catch, they can remain internal. I did several try/error with my test and forgot to loop back to hide those.
I've implemented the suggested change.

internal/install/recipe_installer_test.go
Julien4218: Yes we do want to fail, this test has been moved to os_validator_test.go

internal/install/discovery/os_windows_validator.go
Julien4218: I'm not quite onboard with this, the main reason I find string easier to test with on the unit test side instead of adding several error.Error() calls to get back to the strings which makes the tests a lot harder to read.

In the end, a final error is raised after concatenating all validation messages into 1 string. 

Side note, the smaller validators (like OsValidator) are not meant to be used outside of the discovery package. Also, those validators are not erroring to do their work, they just signal back something potentially useful to the caller.

internal/install/discovery/os_windows_validator.go
```diff
@@ -0,0 +1,62 @@
...
+	if m.OS != "windows" {
+		return ""
+	}
+
+	versions := strings.Split(m.PlatformVersion, ".")
```
sanderblue, line 28: Not a blocker by any means, but the [`go-semvery`](https://github.com/coreos/go-semver) library has a nice API for these kinds of checks for things like `ensureMinimumVersion`.

internal/install/discovery/mock_discoverer.go
```diff
@@ -24,6 +25,14 @@ func (d *MockDiscoverer) Os(os string) {
 	d.DiscoveryManifest.OS = os
 }
 
+func (d *MockDiscoverer) PlatformVersion(pf string) {
```
sanderblue, line 28: Should we name this `SetPlatformVersion`?

Julien4218: I don't know, the other method "Os" didn't have a "SetOs", so I kept consistency. I kind of like for this type of test object to behave like a builder, and it makes it a bit more readable without the Set?

internal/install/discovery/os_windows_validator.go
Julien4218: Good to know, I did struggle a bit with the golang regexp library yesterday, to the point I realize I didn't need a library to split on a "." :) 
But yeah, good to know about this library I guess.

internal/install/discovery/os_windows_validator.go
sanderblue: Agree for a couple reasons - returning an error (or `nil`) follows the patterns we've set forth in this repo which follow Go standards to the best of our knowledge and returning an error provides the ability to return typed errors, so we could potentially have custom error types for these error scenarios.

internal/install/discovery/os_windows_validator.go
zlesnr: Can you point out which test is simpler by returning a string here?  If you use the concrete types of an error, you don't have to check the `error.Error()` method, you can check the error type itself.

zlesnr: Here is an example of using the `require` package in a unit test that compares two error types.  https://pkg.go.dev/github.com/stretchr/testify/require#Error

Julien4218: Some impacted usages below.

It seems a few folks are lining up behind the usage of error versus string. I kind of wish go was having a nullable string type. To please the quorum, I'll go ahead and update, but makes me certainly prefer other languages. 

image

image

internal/install/discovery/mock_discoverer.go
jbeveland27: I actually had the same thought as I came across this and paused wondering if it should be `setPaltformVersion`. I'm not sure what's idiomatic w.r.t. Go when it comes to setters.

internal/install/discovery/os_windows_validator.go
zlesnr: This might also be of interest for you.  https://pkg.go.dev/github.com/pkg/errors#hdr-Adding_context_to_an_error

internal/install/discovery/mock_discoverer.go
sanderblue: Meh, not gonna nitpick about it, but in my opinion a function call of `PlatformVersion()` semantically feels like it would return me the current platform version and not set it. 

i.e. `os.PlatformVersion()` vs `os.SetPlatformVersion()` have semantically different meanings to me

Either way, not gonna make it a blocker for the PR.

sanderblue (approved):

internal/install/discovery/mock_discoverer.go
Julien4218: I ve updated to a SetXxx semantic

sanderblue (approved):

PR #800, opened 2021-04-13
chore(install): remove entity link when no install successful

Julien4218:


Julien4218:
https://newrelic.atlassian.net/browse/VIRTUOSO-549

internal/install/execution/install_status.go
```diff
@@ -227,6 +227,16 @@ func (s *InstallStatus) isCanceled() bool {
 	return false
 }
 
+func (s *InstallStatus) hasAnyRecipeStatus(status RecipeStatusType) bool {
```
zlesnr, line 230: It looks like the other `has*()` methods could be replaced with this implementation.  Perhaps consider updating those methods as well.

zlesnr (dismissed):

internal/install/execution/install_status.go
Julien4218: Excellent, I was hesitating converting those other usages too, I've replaced as suggested.

sanderblue (approved):

zlesnr (approved):

PR #804, opened 2021-04-14
feat(install): add skip apm option and skip any APM recipe when set

Julien4218:
https://newrelic.atlassian.net/browse/VIRTUOSO-679

### DIFF
--- a/internal/install/types/recipe.go
+++ b/internal/install/types/recipe.go
@@ -1,5 +1,7 @@
 package types
 
+import "strings"
+
 type Recipe struct {
 	ID                string                                   `json:"id"`
 	Description       string                                   `json:"description"`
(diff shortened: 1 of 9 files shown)


internal/install/types/recipe.go
```diff
@@ -68,6 +74,18 @@ func (r *Recipe) HasApplicationTargetType() bool {
 	return r.HasTargetType(OpenInstallationTargetTypeTypes.APPLICATION)
 }
 
+func (r *Recipe) HasKeyword(keyword string) bool {
+	if len(r.Keywords) == 0 {
+		return false
+	}
+
+	for _, single := range r.Keywords {
+		return strings.EqualFold(single, keyword)
```
zlesnr, line 83: This looks like it would only check the first keyword.

zlesnr (changes requested):

internal/install/types/recipe.go
Julien4218: Good catch, thanks @zlesnr

zlesnr (approved):

Issue #803, opened 2021-04-13
CLI outputs warning messages when there is no HOME directory


rittneje:
We run the CLI via a docker container. Since the user has no HOME directory, we get the following warning message:

> level=warning msg="Could not create log file folder: stat //.newrelic: no such file or directory"

One issue in play is that this application is using string concatenation instead of `filepath.Join` in several places, resulting in that double slash.

The other is that I don't understand why it's trying to create a log file. It's not clear from a cursory look through this code why it needs a log file or what it's used for. I would expect all the logs to just go to stderr/stdout (which they are anyway).

zlesnr:
This is good feedback.  Thank you for the report.

zlesnr:
It looks like we try to create the log directory when we initialize the logging configuration.  This was part of the requirement for the `install` command that we are using as part of the "guided installation", but I don't think this should be necessary or useful for the rest of the usage of the CLI.  I'll run this past the team and get some feedback about how we want to proceed.

In the meantime, could you provide the full command that you are running to produce this, omitting any sensitive detail?

rittneje:
We are specifically running the command `newrelic reporting junit -a <account id> -p <results.xml>`. The `NEW_RELIC_API_KEY` and `NEW_RELIC_INSIGHTS_INSERT_KEY` environment variables are also set.

zlesnr:
This will go out with the next release of the CLI.  Let us know if there are any other issues once released.

rittneje:
Can confirm it is working as expected now (no warning logs). Thanks!

PR #805, opened 2021-04-14
fix: init the file logger only for the install command

zlesnr:
fixes: #803

Without this change, we initialize the file logger even when we might not intend
to.  Here we ensure that the file logger is only initialized during execution of
the install command.


jbeveland27 (approved):
🚢

PR #807, opened 2021-04-15
fix(install): allow apm recipes in guided install

Julien4218:


Julien4218:
https://newrelic.atlassian.net/browse/VIRTUOSO-674

internal/install/recipe_installer_guided.go
```diff
@@ -274,7 +274,7 @@ func (i *RecipeInstaller) removeRecipes(recipes []types.Recipe, remove ...types.
 func (i *RecipeInstaller) filterIntegrations(recommendedIntegrations []types.Recipe) ([]types.Recipe, error) {
 	installCandidates := []types.Recipe{}
 	for _, r := range recommendedIntegrations {
-		if r.HasApplicationTargetType() {
+		if r.HasApplicationTargetType() && !r.IsApm() {
```
jbeveland27, line 277: Just confirming here - we now have an agreed upon contract where putting the `apm` keyword in `type: application` recipes will result in that being run for Guided Install, correct? If so, would like to document that somewhere (maybe in our internal docs repo).

jbeveland27 (approved):
LGTM 🚀

internal/install/recipe_installer_guided.go
Julien4218: You're correct @jbeveland27 
This is an intermediate situation and once we have more clarity regarding APM support in guided install (only .net for now), and perhaps once the CLI has a more direct access to the recipe, we could re-visit the recipe metadata (need a field to identify guided?)

Julien4218: @jbeveland27 I forgot, where would you suggest to keep the documentation you mentioned?

jbeveland27: Actually might be best to just put a doc (or add to an existing one) in the open-install-library repo: https://github.com/newrelic/open-install-library/tree/main/docs

Julien4218: @jbeveland27 how is this https://github.com/newrelic/open-install-library/commit/9d2d24162ffb8882acac967591bc337ce696b982

jbeveland27: Yup that looks good!

PR #821, opened 2021-04-27
feat(recipes): introduce new flag to use local recipe directory

zlesnr:


internal/install/types/discovery_manifest.go
```diff
@@ -29,3 +35,61 @@ type MatchedProcess struct {
 func (d *DiscoveryManifest) AddMatchedProcess(p MatchedProcess) {
 	d.Processes = append(d.Processes, p)
 }
+
+func (d *DiscoveryManifest) ConstrainRecipes(allRecipes []Recipe) []Recipe {
```
zlesnr, line 39: Double check my logic here to make sure this is what we want.

internal/install/recipes/errors.go
```diff
@@ -0,0 +1,6 @@
+package recipes
+
+import "github.com/pkg/errors"
+
+// ErrRecipeNotFound is used when a recipe is requested by name, but does not exist for the given constraint.
+var ErrRecipeNotFound = errors.New("recipe not found")
```
sanderblue, line 6: 🥇

internal/install/recipes/local_recipe_fetcher.go
```diff
@@ -0,0 +1,112 @@
...
+	log.WithFields(log.Fields{
+		"path": path,
+	}).Debug("loading recipes")
+
+	re := regexp.MustCompile(`\.ya?ml`)
+
+	err := filepath.Walk(
+		path,
+		func(path string, info os.FileInfo, err error) error {
+			if re.MatchString(path) {
```
sanderblue, line 72: Not that it's a big deal, but would checking the file's extension directly with Go using `filepath.Ext(path)` make things "safer"? (i.e. "safer" for devs and inline readability and not relying on the wonderful mysterious world of regex 🦙 😉 ).

internal/install/types/discovery_manifest.go
```diff
@@ -29,3 +35,57 @@ type MatchedProcess struct {
...
+				if !strings.EqualFold(target.PlatformVersion, d.PlatformVersion) {
+					continue
+				}
+			}
+
+			recipes = append(recipes, recipe)
+		}
+	}
+
+	log.Debugf("%d embedded recipes found for manifest", len(recipes))
```
sanderblue, line 88: Nitpick, do we want to say `embedded`?

zlesnr: Oh good call.  I think that was a copy-paste from an earlier version.

sanderblue (approved):
🐑 it

Issue #774, opened 2021-03-24
Missing support for ARM processors


cwinters8:
TLDR: Are there any plans to support ARM based processors, such as the new-ish AWS Graviton 2 processors? 

I specifically want to be able to monitor my MongoDB infrastructure using New Relic, but it seems there isn't any support for ARM in either the CLI or in the New Relic Infra MongoDB integration. It seems like my options are either change instance types (don't want to do - the m6g.medium EC2 instance type is just right for our use case currently, and going to an m5 instance means significantly higher expense) or switch to DataDog (would prefer to use NR). Please let me know if I'm missing an option. Thanks in advance.

### Description
Cannot install on ARM based Amazon Linux instances

### Current behavior
Attempting to install using the following command:
```bash
curl -Ls https://raw.githubusercontent.com/newrelic/newrelic-cli/master/scripts/install.sh | bash && sudo NEW_RELIC_API_KEY=REDACTED NEW_RELIC_ACCOUNT_ID=3112141 /usr/local/bin/newrelic install -n mongodb-open-source-integration
```
Results in:
```
Starting installation.
Installing New Relic CLI v0.20.3
Installing to /usr/local/bin using sudo
/usr/local/bin/newrelic: /usr/local/bin/newrelic: cannot execute binary file
```
Based on some additional investigation in the releases for newrelic-cli, it looks like only x86 architectures are supported currently.

### Expected behavior
I would expect the CLI to be able to execute after installation.

jthurman42:
There shouldn't be anything specific in the CLI that breaks on ARM, this is likely more of an issue that the build system does not generate an artifact (currently only `amd64` is build: https://github.com/newrelic/newrelic-cli/blob/master/.goreleaser.yml#L21)

Each integration on the other hand would need to be tested / build / updated as well.

jthurman42:
Tested goreleaser with the added `arm64` field, and there were no compilation issues.  I did not spin up an arm64 host to test the binary after the fact.  ( https://github.com/newrelic/newrelic-cli/blob/master/.goreleaser.yml#L21 )

Also note that the install script above would need ARCH detection, it's hard-coded at `X86_64`: https://github.com/newrelic/newrelic-cli/blob/master/scripts/install.sh#L50

cwinters8:
Awesome, thank you @jthurman42. I will give that a shot.

Also, look what appeared today for nri-mongodb!! I'm so happy!
[...]

PR #828, opened 2021-04-29
feat(build): add arm64 support

zlesnr:


zlesnr:
Relates to #774

scripts/install.sh
```diff
@@ -22,6 +22,18 @@ else
 	exit 1
 fi
 
+
+if [ "$(uname -m)" == "x86_64" ]; then
+    MACHINE="x86_64"
```
sanderblue, line 27: Should we set this variable at the top of the file with `x86_64` as the default? Seems to be a prevalent pattern in the make files.

scripts/install.sh
```diff
@@ -22,6 +22,18 @@ else
...
+elif [ "$(uname)" == "aarch64" ]; then
+    MACHINE="aarch64"
+else
+	echo "This machine architecture is not supported."
```
sanderblue, line 31: Maybe we list the supported architectures for clarity? 

> "This machine architecture is not supported. The supported architectures are x86_64 and aarch64."

scripts/install.sh
```diff
@@ -14,12 +14,21 @@ echo "Starting installation."
...
+    exit 1
+fi
+
+if [ "$(uname -m)" == "x86_64" ]; then
+    MACHINE="x86_64"
+elif [ "$(uname)" == "aarch64" ]; then
+    MACHINE="aarch64"
+else
+    echo "This machine architecture is not supported. The supported architectures are x86_64 and aarch64."
+    exit 1
```
jbeveland27, line 31: Just curious - we're building for arm 6/7, but not looking for that machine type here (or at least I don't think we are). Should this be adjusted will this work and I'm just missing something?

zlesnr: Yeah, I want to see what the archive names are and then I can update the script.

jbeveland27: Or should we just not build for 32-bit architectures?

zlesnr: But that will take a release.

jbeveland27: Cool, makes sense

jbeveland27 (approved):

PR #829, opened 2021-04-30
chore(install): make go-task errors nicer

ctrombley:
This PR has a few effects:
1. Removes a bunch of the noise from go-task errors
2. Appends the final line of stderr to the go-task error so that it will appear in the final FATAL message presented to the user

### Testing
```
newrelic installTest -c test/recipes/testError.yaml
```

The `testError.yaml` file can be updated as desired to try various recipe invocation scenarios.

### Notes

1. Removing the "noise" in the error also loses some of the context about which sub-task in the recipe the error originated from. (e.g. `task: Failed to run task "default": task: Failed to run task "execute":`) Is this acceptable?
1. Errors will show up in the `FATAL` message if:
   1. The convention `echo -e "error message" >&2` is used directly in the recipe.  (`echo -e "eror message" > /dev/null` **will not** work since that routes output around our capture buffer and directly to the system level stderr file handler.  This new convention will likely require existing recipes to be updated to make use of the new feature.)
   1. A child process invoked by the recipe fails, *and* provides an error message as the last line of its stderr output.  This can sometimes be confusing if the error from the child process is split across multiple lines.  To demonstrate this, you can pass a malformed option to `jq` (I used `jq -asdf`) and watch what happens:

```
==> Installing test-error...
jq: Unknown option -asdf
Use jq --help for help with command-line options,
or see the jq manpage, or online docs  at https://stedolan.github.io/jq
==> Installing test-error...failed.

WARNING encountered an error while executing test-error: exit status 2: or see the jq manpage, or online docs  at https://stedolan.github.io/jq 
WARNING execution of Test error failed, please see the following link for clues on how to resolve the issue: https://docs.newrelic.com/docs/using-new-relic/cross-product-functions/troubleshooting/not-seeing-data/ 
  One or more installations failed.  Check the install log for more details: /Users/ctrombley/.newrelic/newrelic-cli.log
  New Relic installation complete!


FATAL test failed: encountered an error while executing test-error: exit status 2: or see the jq manpage, or online docs  at https://stedolan.github.io/jq 
```

I'm not sure what do do about this, but my thinking is that more information is better than none.  I am curious to know if the team feels that this caveat invalidates the approach.  Suggestions are invited.

### DIFF
--- a/internal/install/types/errors.go
+++ b/internal/install/types/errors.go
@@ -12,6 +12,7 @@ var ErrInterrupt = errors.New("operation canceled")
 type GoTaskError interface {
 	error
 	Tasks() []string
+	SetError(msg string)
 }
 
 // GoTaskError represents a task failure reported by go-task.
@@ -22,6 +23,10 @@ type GoTaskGeneralError struct {
 
 func (e GoTaskGeneralError) Error() string {
 	return e.error.Error()
+}
+
+func (e GoTaskGeneralError) SetError(msg string) {
+	e.error = errors.New(msg)
 }
 
 func (e GoTaskGeneralError) Tasks() []string {
@@ -52,7 +57,7 @@ type ErrNonZeroExitCode struct {
 	additionalContext string
 }
 
-func NewNonZeroExitCode(originalError GoTaskGeneralError, additionalContext string) ErrNonZeroExitCode {
+func NewNonZeroExitCode(originalError GoTaskError, additionalContext string) ErrNonZeroExitCode {
 	return ErrNonZeroExitCode{
 		GoTaskError:       originalError,
 		additionalContext: additionalContext,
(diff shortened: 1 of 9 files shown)


Julien4218 (approved):
Interesting work, we might want to do a bit of testing across the recipes (with bogus input to trigger errors) to check the output and what shows in Tessen

internal/install/types/errors.go
```diff
@@ -2,24 +2,65 @@ package types
...
+
+	var task string
+	if len(parsed) > 0 && len(parsed[0]) > 0 {
+		task = string(parsed[0][1])
+	}
+
+	stripped := re.ReplaceAllString(err.Error(), "")
+
+	return GoTaskGeneralError{
+		tasks: []string{task},
```
gabeobrien, line 44: Does 'tasks' hold more than 1 task?  It looks like the code is setup to find 1 and only 1 task.  Maybe we should call it just 'task'?

ctrombley: This involves a change to the contract between the front and back end. I wanted to leave the type open to defining a variable length path to support arbitrary task nesting, but didn't implement just yet.

zlesnr (dismissed):

Julien4218 (approved):

Julien4218 (approved):

PR #842, opened 2021-05-10
chore(install): provide redirect URL on nerdstorage doc for frontend to use

sanderblue:
Currently we have very similar logic on the frontend as we do in the CLI. This PR ensures a backwards compatible way to provide a redirect URL for the frontend to use when generating the link used for "See Your Data" upon a successful installation.

### DIFF
--- a/internal/install/execution/terminal_reporter.go
+++ b/internal/install/execution/terminal_reporter.go
@@ -2,22 +2,17 @@ package execution
 
 import (
 	"fmt"
-	"strings"
 
 	log "github.com/sirupsen/logrus"
 
 	"github.com/newrelic/newrelic-cli/internal/install/types"
 )
 
-type TerminalStatusReporter struct {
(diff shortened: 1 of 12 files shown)


ctrombley (approved):

internal/install/execution/terminal_reporter.go
```diff
@@ -98,6 +98,10 @@ func (r TerminalStatusReporter) InstallComplete(status *InstallStatus) error {
 
 	fmt.Println("  New Relic installation complete!")
 
+	log.Print("\n****************************\n")
```
ctrombley, line 101: Left this here by mistake?

sanderblue: dang it, yes

zlesnr (dismissed):

ctrombley (approved):
LGTM

PR #843, opened 2021-05-11
fix(install): replace AddVar() with SetRecipeVar() after types refactor

sanderblue:
Missed this update due to context switching in the middle the large types refactor PR 🦙 😎


Julien4218 (approved):

jbeveland27 (approved):
🐑

PR #844, opened 2021-05-12
fix(chore): add validation rule for min version of Ubuntu and Windows

Julien4218:
https://newrelic.atlassian.net/browse/VIRTUOSO-885


jlegoff (approved):
LGTM

PR #845, opened 2021-05-13
(feat): validate CLI config

ctrombley:
Depends on newrelic/newrelic-client-go#724.

### Testing
Run cold start command from a fresh VM without the CLI installed.  Once it is finished:

```
sudo newrelic profile list --show-keys
```

You should see that the Insights insert key should have been retrieved and populated when the profile was created.  

Then,

```
sudo newrelic diagnose validate
```

The validation process will kick off and report success once it finds the event.  To locate the diagnostic event manually, run the following query:

```
SELECT * FROM NrIntegrationError WHERE purpose = 'New Relic CLI configuration validation'
```

### To do (future PRs)

- Improve success and error messages
- Improve `profile add` command to make use of the key location functionality

### DIFF
--- a/cmd/newrelic/command.go
+++ b/cmd/newrelic/command.go
@@ -177,11 +177,6 @@ func fetchLicenseKey(client *newrelic.NewRelic, accountID int) (string, error) {
 	return "", types.ErrorFetchingLicenseKey
 }
 
-type insightsKey struct {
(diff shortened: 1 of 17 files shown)


sanderblue (approved):
💲 💵 :heavy_dollar_sign:

PR #854, opened 2021-05-18
feat(install): Add account scoped nerd storage write

murphpdx:
The PLG team is doing some work around recommendations. Our team is working on a feature that would display all recommendations for a specific user.

Problem we would like to solve:
The way the recommendations are currently stored will make getting the recommendations extremely costly. Our team would need to loop through every account the user has access to then every entity for every account and check for documents in Nerd Store.

Proposed Solution:
We would like to store the same documents in account scoped nerdstore. This solution allows us to then get all accounts the user has access to and go right to Nerd Store to check for documents.

There are some limitations to this solution. We understand that we may run into nerd store limits but based on the total number of these documents currently stored in production (~57k) we think this solution will be sufficient for now. Eventually, we would like to implement something more robust.


ctrombley (approved):
🚢

PR #868, opened 2021-05-20
fix(install): UnmarshalYAML toStringByFieldName() should handle multiple types when converting interface 

sanderblue:
This PR updates the `OpenInstallationRecipe` custom `UnmarshalYAML()` method to handle converting multiple types to a string.

I ran into panic errors when unmarshaling the elasticsearch recipe via `--recipePath` because the input variables of the port and SSL checks are `int` and `bool` respectively. The `toStringByFieldName()` helper function needed to be updated to handle these scenarios and others like it.

### DIFF
--- a/internal/install/types/recipe.go
+++ b/internal/install/types/recipe.go
@@ -2,6 +2,7 @@ package types
 
 import (
 	"fmt"
+	"strconv"
 	"strings"
 
 	"gopkg.in/yaml.v2"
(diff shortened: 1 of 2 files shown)


internal/install/types/recipe.go
```diff
@@ -284,11 +285,20 @@ func toBoolByFieldName(fieldName string, data map[string]interface{}) bool {
...
 func toStringByFieldName(fieldName string, data map[string]interface{}) string {
-	if v, ok := data[fieldName]; ok {
-		return v.(string)
+	out := ""
+	if in, ok := data[fieldName]; ok {
+		switch v := in.(type) {
+		case int:
+			return strconv.Itoa(v)
+		case string:
+			return v
```
jbeveland27, line 294: Noob Go question - does `return v.(string)` make sense here or is the type assertion not as valuable because you're already switching on type?

jbeveland27 (approved):
Nice catch 🚢

internal/install/types/recipe.go
sanderblue: The first part of the switch statement `v := in.(type)` makes `v` a string so the type assertion is not necessary in that block. This is kind of unique to [`switch` by type](https://tour.golang.org/methods/16) in Go.

jbeveland27: Ooooh that's nice

PR #870, opened 2021-05-21
fix(diagnose): increase retry and move output to the last attempt

Julien4218:


Julien4218:
@ctrombley I noticed we needed to increase the retry for freshly created new accounts (it took about 15 times for a new account on infra windows).

ctrombley (approved):

sanderblue (approved):

PR #873, opened 2021-05-24
Correctly detect aarch64

ruffy91:
The installation on aarch64 did not work, because the detection logic was missing a parameter.
Added the -m to uname to correctly get the machine architecture


ruffy91:
I still made an error and will shortly reopen a new PR

PR #874, opened 2021-05-24
Update install.sh to work on ARM64

ruffy91:
The installation on aarch64 did not work because uname was missing the -m parameter.
Also the release artifact for ARM64 is called arm64 and not aarch64.

I corrected both and tested it on a Ubuntu 20.04 installation on a Raspberry Pi 4B.


ctrombley:
Thanks @ruffy91 !

PR #879, opened 2021-05-25
fix(install): targeted install uses process detection

noahmmcgivern:


Julien4218:
@noahmmcgivern can you get latest from the main branch, there are a couple of files conflicting.
Also lint is unhappy.

PR #883, opened 2021-05-28
feat(install): add UNSUPPORTED status

sanderblue:
**To do**
- [x] Test on different operating systems and use mismatched configs
- [x] Add additional test cases for new code

### DIFF
--- a/internal/install/recipe_installer.go
+++ b/internal/install/recipe_installer.go
@@ -266,6 +266,14 @@ func (i *RecipeInstaller) installRecipes(ctx context.Context, m *types.Discovery
 			log.Debugf("Failed while executing and validating with progress for recipe name %s, detail:%s", r.Name, err)
 			log.Warn(err)
 			log.Warn(i.failMessage(r.DisplayName))
+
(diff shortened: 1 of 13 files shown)


internal/install/recipe_installer.go
```diff
@@ -266,6 +268,19 @@ func (i *RecipeInstaller) installRecipes(ctx context.Context, m *types.Discovery
 			log.Debugf("Failed while executing and validating with progress for recipe name %s, detail:%s", r.Name, err)
 			log.Warn(err)
 			log.Warn(i.failMessage(r.DisplayName))
+
+			if _, ok := err.(*types.UnsupportedOperatingSytemError); ok {
```
Julien4218, line 272: @sanderblue  I dont think we need this logic block, since we already have a return with this logic on line 262?

sanderblue: Will double check. Thanks

sanderblue: Hmm, I think you're right. Testing this it out now.

Julien4218 (approved):

PR #889, opened 2021-06-03
Refactor/simplify recipe install

ctrombley:


internal/install/recipes/recipe_filterer.go
```diff
@@ -0,0 +1,175 @@
...
+func NewProcessMatchRecipeFilter() *ProcessMatchRecipeFilter {
+	return &ProcessMatchRecipeFilter{
+		processMatchFinder: NewRegexProcessMatchFinder(),
+	}
+}
+
+func (f *ProcessMatchRecipeFilter) Execute(ctx context.Context, r *types.OpenInstallationRecipe, m *types.DiscoveryManifest) (bool, error) {
+	matches, err := f.processMatchFinder.FindMatches(ctx, m.DiscoveredProcesses, *r)
+	if err != nil {
+		return false, err
```
Julien4218, line 106: @ctrombley similarly to the ScriptEvaluationRecipeFilter, if we error while trying to see if we can filter on process match, I believe we should: 
- display a debug/warn/error message 
- return `true` for filtered.

internal/install/recipe_installer.go
```diff
@@ -159,17 +171,48 @@ func (i *RecipeInstaller) discoverAndRun(ctx context.Context) error {
...
+	candidateRecipes := append(recipesForPlatform, targetedRecipes...)
+	filteredRecipes, err := i.recipeFilterer.FilterMultiple(ctx, candidateRecipes, m)
+	if err != nil {
+		return err
+	}
+	log.Tracef("recipes after filtering: %v\n", filteredRecipes)
+
+	orderedAndDeduped := orderAndDedupeRecipes(filteredRecipes)
+
+	selected, unselected, err := i.promptUserSelect(orderedAndDeduped)
```
Julien4218, line 195: I don't think we want to prompt for Targeted

internal/install/recipe_installer.go
```diff
@@ -316,31 +364,142 @@ func (i *RecipeInstaller) failMessage(componentName string) error {
...
+}
+
+func orderAndDedupeRecipes(recipes []types.OpenInstallationRecipe) []types.OpenInstallationRecipe {
+	deduped := dedupeRecipes(recipes)
+	ordered := orderRecipes(deduped)
+
+	return ordered
+}
+
+func dedupeRecipes(recipes []types.OpenInstallationRecipe) []types.OpenInstallationRecipe {
```
Julien4218, line 476: The dedup logic is complex. 
We need to find the best match looking at the recipe installTarget and discoveryManifest. 
In case of multiple findings, the most matching one should be returned. Any candidate not matching any of the recipe criteria should be discarded.

PR #894, opened 2021-06-09
Fix/recipe host matching

Julien4218:


Julien4218:
https://newrelic.atlassian.net/browse/VIRTUOSO-1053

internal/install/recipes/recipe_repository.go
```diff
@@ -117,9 +117,16 @@ func findMaxMatch(matches []recipeMatch) recipeMatch {
...
 	for _, match := range matches {
 		if result == nil {
-			result = &match
+			result = &recipeMatch{
+				recipe:     match.recipe,
+				matchCount: match.matchCount,
+			}
 		} else {
 			if match.matchCount > result.matchCount {
+				result = &recipeMatch{
```
ctrombley, line 126: This assignment is unnecessary is we're setting `result` again on line 130, right?

Julien4218: yep, good catch.

Julien4218: fixed

ctrombley (approved):

PR #901, opened 2021-06-10
feat(install): add observability pack creation to install flow

jbeveland27:
# ~DO NOT MERGE!~ Ready for Review

~Need to make sure `observabilityPackSearch` is generally available in NerdGraph before merging (currently behind feature flag), otherwise the calls for fetching Observability Packs will fail. However, until `observabilityPacks` fields are added to recipes, this _shouldn't_ cause any harm. However out of caution it's probably best to ensure the schema is generally available so as to not cause installation issues.~

`observabilityPackSearch` is now generally available in NerdGraph, so this PR can be merged.


internal/install/packs/packs_installer.go
```diff
@@ -0,0 +1,136 @@
...
+	// Check for existence of dashboard before creating a new one
+	dashboards, err := p.client.Dashboards.ListDashboards(&dashboards.ListDashboardsParams{
+		Title: dashboard.Name,
+	})
+	if err != nil {
+		return nil, fmt.Errorf("error checking if dashboard already exists: %s", err)
+	}
+
+	if len(dashboards) > 0 {
+		fmt.Printf("  ==> Dashboard [%s] already exists, skipping\n", dashboard.Name)
```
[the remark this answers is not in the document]
jbeveland27: This was more so to indicate a "progress" for installing the parts of a pack (and would be added to in the future when we start installing alerts, synthetics, etc). Though there's not really any hard reqs around the UX for this feature, so I can make it a debug if you'd still prefer.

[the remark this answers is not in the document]
zlesnr: Feedback to the user makes sense, and the formatting you have there looks good also.

internal/install/packs/packs_installer.go
```diff
@@ -0,0 +1,136 @@
...
+	defer func() { p.progressIndicator.Stop() }()
+
+	for _, pack := range packs {
+		fmt.Printf("\n  Installing pack: %s\n", pack.Name)
+
+		// Only installing dashboards currently
+		if pack.Dashboards != nil {
+			for _, dashboard := range pack.Dashboards {
+				if _, err := p.createObservabilityPackDashboard(ctx, dashboard); err != nil {
+					return fmt.Errorf("Failed to create observability pack dashboard [%s]: %s", dashboard.Name, err) // nolint: golint
```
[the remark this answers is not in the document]
jbeveland27: Good call. Pushed an update.

internal/install/packs/packs_installer.go
```diff
@@ -0,0 +1,137 @@
...
+	"github.com/newrelic/newrelic-client-go/newrelic"
+	"github.com/newrelic/newrelic-client-go/pkg/dashboards"
+	"github.com/newrelic/newrelic-client-go/pkg/entities"
+	log "github.com/sirupsen/logrus"
+)
+
+// Maybe both fetchPacks and installPacks should be one interface
+// vs separate interfaces?
+// nolint: golint
+type PacksInstaller interface {
```
ctrombley, line 23: Separate interfaces is fine, you can have a single concrete implementation that conforms to both if you like.  There is also now an `interfaces.go` that allows defining the interfaces closer to the usage (ie `install` package).  This will allow you to get rid of the `Concrete` prefix on the concrete type.

internal/install/packs/packs_installer.go
```diff
@@ -0,0 +1,137 @@
...
+
+	for _, pack := range packs {
+		fmt.Printf("\n  Installing pack: %s\n", pack.Name)
+
+		// Only installing dashboards currently
+		if pack.Dashboards != nil {
+			for _, dashboard := range pack.Dashboards {
+				if _, err := p.createObservabilityPackDashboard(ctx, dashboard); err != nil {
+					p.progressIndicator.Fail(msg)
+					return fmt.Errorf("Failed to create observability pack dashboard [%s]: %s", dashboard.Name, err) // nolint: golint
```
ctrombley, line 56: The linter will complain if the error message starts with a capitalized letter.

internal/install/packs/packs_installer.go
```diff
@@ -0,0 +1,137 @@
...
+
+	dashboard, err := transformDashboardJSON(body, defaultProfile.AccountID)
+	if err != nil {
+		return nil, err
+	}
+
+	fmt.Printf("  ==> Creating dashboard: %s\n", dashboard.Name)
+
+	// Check for existence of dashboard before creating a new one
+	dashboards, err := p.client.Dashboards.ListDashboards(&dashboards.ListDashboardsParams{
```
ctrombley, line 83: This will break when the Insights EOL happpens.  There is no Nerdgraph-based replacement in the client yet.  I'll create an issue for that work.

internal/install/packs/packs_installer.go
```diff
@@ -0,0 +1,137 @@
...
+		return nil, fmt.Errorf("error checking if dashboard already exists: %s", err)
+	}
+
+	if len(dashboards) > 0 {
+		fmt.Printf("  ==> Dashboard [%s] already exists, skipping\n", dashboard.Name)
+		return nil, nil
+	}
+
+	// Dashboard doesn't exist yet, proceed with dashboard create
+	created, err := p.client.Dashboards.DashboardCreate(accountID, dashboard)
```
ctrombley, line 96: This method is deprecated in the client.  Use `DashbordCreateWithContext`.

internal/install/packs/packs_installer.go
```diff
@@ -0,0 +1,137 @@
...
+
+	p.progressIndicator.Success(msg)
+	return nil
+}
+
+func (p *ConcretePacksInstaller) createObservabilityPackDashboard(ctx context.Context, d types.OpenInstallationObservabilityPackDashboard) (*dashboards.DashboardCreateResult, error) {
+	defaultProfile := credentials.DefaultProfile()
+	accountID := defaultProfile.AccountID
+
+	body, err := getJSONfromURL(d.URL)
```
ctrombley, line 70: This should ideally be implemented in the client, with types and potentially API methods generated by tutone.  Create an issue for the work?

internal/install/packs/service_packs_fetcher.go
```diff
@@ -0,0 +1,111 @@
...
+		client: client,
+	}
+
+	return &f
+}
+
+func (f *ServicePacksFetcher) FetchPacks(ctx context.Context, recipes []types.OpenInstallationRecipe) ([]types.OpenInstallationObservabilityPack, error) {
+	log.Tracef("FetchPacks.recipes: %+v", recipes)
+
+	packs := []types.OpenInstallationObservabilityPack{}
```
ctrombley, line 29: Should this be generated by tutone and live in the client?

internal/install/packs/packs_installer.go
jbeveland27: Yeah, that's why I added the `// nolint: golint` to the end. Is it more idiomatic/preferred to just not capitalize the error?

internal/install/packs/packs_installer.go
jbeveland27: Do you have a link? I can't seem to locate that method.

internal/install/packs/packs_installer.go
jbeveland27: Agreed that we should do this eventually. From last meeting, sounds like we're going to punt on this for now.

internal/install/packs/service_packs_fetcher.go
jbeveland27: Agreed that we should do this eventually. From last meeting, sounds like we're going to punt on this for now.

internal/install/packs/packs_installer.go
jbeveland27: Issue created: https://github.com/newrelic/newrelic-client-go/issues/748

internal/install/packs/service_packs_fetcher.go
jbeveland27: Issue created: https://github.com/newrelic/newrelic-client-go/issues/748

internal/install/packs/packs_installer.go
jbeveland27: Cool, I knew a single implementation could fulfill both interfaces somewhere in the back of my mind, just didn't think to do that. I kept the implementations separate for now just because it felt cleaner separating the concerns, but did shuffle the interface around and renamed `ConcretePacksInstaller` to `ServicePacksInstaller`. Moving the interfaces to `interfaces.go` displaces those outside of the `packs` package, and that seemed odd (plus the compiler was yelling at me and I couldn't quite figure out how to get it to resolve short of getting rid of the `packs` directory altogether).

ctrombley: Happy to pair on this if you like, maybe I'm just missing something.

internal/install/packs/packs_installer.go
ctrombley: Ah, I didn't see the hint. So far we've taken the approach of letting the linter have its way unless there's a good reason not to, but I don't think anyone is going to have an especially strong opinion about this.

internal/install/packs/packs_installer.go
ctrombley: https://github.com/newrelic/newrelic-client-go/blob/main/pkg/dashboards/dashboards_api.go#L22-L39

jbeveland27: Ohhh gotcha. Looks like that was just added a few days ago which is why I was missing it. Thanks.

internal/install/packs/packs_installer.go
jbeveland27: Gotcha. I updated to conform to the linter overload 🙇🏼

ctrombley (approved):
🌮

ctrombley (dismissed):

devfreddy (approved):

PR #905, opened 2021-06-10
chore(install): let proxy users know when an endpoint is unreachable

zlesnr:


internal/install/recipe_installer.go
```diff
@@ -522,3 +523,18 @@ func addIfMissing(recipes []types.OpenInstallationRecipe, dependencies []types.O
...
+
+func checkNetwork(nrClient *newrelic.NewRelic) {
+	err := nrClient.TestEndpoints()
+	if err != nil {
+
+		proxyConfig := httpproxy.FromEnvironment()
+
+		log.Debugf("proxyConfig: %+v", proxyConfig)
+		if proxyConfig.HTTPProxy != "" || proxyConfig.HTTPSProxy != "" || proxyConfig.NoProxy != "" {
+			log.Warn("You may need to adjust your proxy environment variables, ie: HTTPS_PROXY")
```
jbeveland27, line 535: Should we link somewhere that tells the user how to set proxy vars?

zlesnr: Good question.  Will the UI have checkboxes for enabling this?  @michaelkro do we know what this will look like on the front end for how a user will enable the proxy to change the environment variables?

zlesnr: Good link.  I've updated the message.

jbeveland27 (dismissed):
LGTM 🚢 🇮🇹

internal/install/recipe_installer.go
michaelkro: Whoops, so sorry, I totally forgot to respond.

Yes, the user will input the URL and the UI will automatically add it as an env variable, e.g.

```sh
sudo HTTPS_PROXY=https://my-proxy:8000 NEW_RELIC_REGION=STAGING ...
```

devfreddy (approved):
LGTM 🚢

PR #940, opened 2021-06-28
chore: update client to v0.60.1, add context to config validation

ctrombley:
- update `newrelic-client-go` to v0.60.1, including generated `workload` types
- make `utils.Retry` context-aware
- use context-aware methods during config initialization, config validation, and a few other locations

### DIFF
--- a/cmd/newrelic/command.go
+++ b/cmd/newrelic/command.go
@@ -1,11 +1,11 @@
 package main
 
 import (
+	"context"
 	"errors"
 	"fmt"
 	"os"
 	"strconv"
-	"time"
 
 	"github.com/jedib0t/go-pretty/v6/text"
 	log "github.com/sirupsen/logrus"
@@ -37,10 +37,10 @@ var Command = &cobra.Command{
 }
 
 func initializeCLI(cmd *cobra.Command, args []string) {
-	initializeProfile()
+	initializeProfile(utils.SignalCtx)
 }
 
-func initializeProfile() {
+func initializeProfile(ctx context.Context) {
 	var accountID int
 	var region string
 	var licenseKey string
@@ -91,7 +91,7 @@ func initializeProfile() {
 
 			if licenseKey == "" {
 				// We should have an account ID by now, so fetch the license key for it.
-				licenseKey, err = fetchLicenseKey(nrClient, accountID)
+				licenseKey, err = fetchLicenseKey(ctx, nrClient, accountID)
 				if err != nil {
 					log.Error(err)
 					return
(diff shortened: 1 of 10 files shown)


cmd/newrelic/command.go
```diff
@@ -150,30 +150,40 @@ func hasProfileWithDefaultName(profiles map[string]credentials.Profile) bool {
...
+func fetchLicenseKey(ctx context.Context, client *newrelic.NewRelic, accountID int) (string, error) {
 	query := `query($accountId: Int!) { actor { account(id: $accountId) { licenseKey } } }`
 
 	variables := map[string]interface{}{
 		"accountId": accountID,
 	}
 
-	for i := 0; i < 3; i++ {
+	var licenseKey string
+	execLicenseKeyRequest := func() error {
```
Julien4218, line 161: I tripped on the return "" below, only to realize this is a rather long inline function.
Is there a way to break this func out to its own private func, to avoid confusion when reading? just a suggestion.

Julien4218 (dismissed):

cmd/newrelic/command.go
ctrombley: There is not, since it closes over `licenseKey` in order to extract the inner return value.

internal/diagnose/config_validator.go
```diff
@@ -143,7 +143,8 @@ func (c *ConfigValidator) validateLicenseKey(profile *credentials.Profile) error
 		},
 	}
 
-	licenseKeys, err := c.client.APIAccess.SearchAPIAccessKeys(params)
+	fmt.Printf("ctx: %v\n", ctx)
```
jbeveland27, line 146: Maybe consider debug/trace?

ctrombley: This wasn't supposed to survive 😆

internal/install/recipe_installer.go
```diff
@@ -156,7 +156,7 @@ func (i *RecipeInstaller) Install() error {
 		i.status.InstallCanceled()
 		return nil
 	case err = <-errChan:
-		if err == types.ErrInterrupt {
+		if errors.Is(err, types.ErrInterrupt) {
```
jbeveland27, line 159: `ooh_nice.gif`

Julien4218 (dismissed):

jbeveland27 (dismissed):

cmd/newrelic/command.go
ctrombley: I refactored this a bit.

internal/install/recipe_installer.go
ctrombley: There's a bit we can do to make this generally better throughout the project.

jbeveland27 (approved):

PR #948, opened 2021-07-01
re-enable install validation, refactor configuration and profiles modules

ctrombley:
This PR re-enables profile validation for the `install` command, but also contains a significant refactoring of the code used to store configuration and profile (authentication) information.  [`viper`](https://github.com/spf13/viper) has been removed as a dependency since it does not provide [support for case-sensitive keys](https://github.com/spf13/viper/issues/1014), necessary here to allow compatibility with the config file conventions set forth by the NR1 CLI.


Julien4218:
Tested on a few linux instances with many re-install scenarios, all worked very well (no profile created during install).

Julien4218 (dismissed):

Julien4218 (approved):

Julien4218 (approved):

PR #954, opened 2021-07-09
fix: remove region from staging/testing urls

brushknight:
Problem:
`https` is not enabled for region links.


Julien4218:
@brushknight I've replaced this PR by this one https://github.com/newrelic/newrelic-cli/pull/955
I've updated the negative test to ensure they're failing for the good reason, and I've created this PR from this repo branch so that automated tests can run to ensure there is no regression (shouldnt).
I'll deploy this shortly.

PR #965, opened 2021-07-13
Fix/ensure log install

Julien4218:


internal/config/logging.go
```diff
@@ -43,6 +43,10 @@ func InitLogger(logLevel string) {
 	}
 }
 
+func GetDefaultLogFilePath() string {
+	return BasePath + "/" + DefaultLogFile
```
ctrombley, line 47: prefer `filepath.Join()` here

Julien4218: Good catch, I've updated.

ctrombley (approved):

PR #988, opened 2021-07-15
feat: add split.io service

abrunner94:
## Context

This PR adds the ability for the NR team to conduct A/B tests and gradually roll out new features. This is done using Split.io, which we have access to at NR.


internal/split/init.go
```diff
@@ -0,0 +1,29 @@
...
+	SplitService *Service
+)
+
+func Init() {
+	region := configAPI.GetConfigString(config.Region)
+	apiKey := GetAPIKeyByRegion(strings.ToLower(region))
+
+	service, err := NewSplitService(apiKey, region)
+	if err != nil {
+		log.Errorf("could not initialize SplitService: %s\n", err)
```
ctrombley, line 22: This should be `log.Fatalf`, or the method should return an error, unless we are committing to checking for nil during every usage of `SplitService`.  Currently if there is an initialization error, we will experience a panic when using the instance.

internal/split/split_service.go
```diff
@@ -0,0 +1,88 @@
...
+	factory, err := client.NewSplitFactory(apiKey, cfg)
+	if err != nil {
+		log.Errorf("Split SDK init error: %s\n", err)
+		return nil, err
+	}
+
+	client := factory.Client()
+	err = client.BlockUntilReady(10)
+	if err != nil {
+		log.Errorf("Split SDK timeout: %s\n", err)
```
ctrombley, line 47: Consider replacing with `return nil, fmt.Errorf("split SDK timeout: %w\n", err)`

internal/split/split_service.go
```diff
@@ -0,0 +1,88 @@
...
+// in localhost mode as defined in their documentation
+func NewSplitService(apiKey string, region string) (*Service, error) {
+	cfg := conf.Default()
+	if apiKey == "localhost" {
+		cfg.SplitFile = CreateMockSplits()
+	}
+
+	factory, err := client.NewSplitFactory(apiKey, cfg)
+	if err != nil {
+		log.Errorf("Split SDK init error: %s\n", err)
```
ctrombley, line 40: Consider replacing with `return nil, fmt.Errorf("split SDK init error: %w\n", err)`

internal/split/split_service.go
```diff
@@ -0,0 +1,88 @@
...
+}
+
+// Get all treatments from Split.io given a list of splits
+func (s *Service) GetAll(splits []string) map[string]string {
+	log.Debugf("Retrieving treatments for splits: %s", splits)
+	return s.client.Treatments(accountID, splits, nil)
+}
+
+// Creates a temporary file with splits used for unit-testing
+func CreateMockSplits() string {
```
ctrombley, line 66: Can this method be moved into the test file, and can we execute the line above `cfg.SplitFile = CreateMockSplits()`  as part of test setup?

internal/split/split_service.go
```diff
@@ -0,0 +1,88 @@
...
+	configAPI "github.com/newrelic/newrelic-cli/internal/config/api"
+)
+
+// Split.io client API keys
+// These keys are client-facing and are only used to fetch splits.
+// There is no security risk in exposing these keys, as the only purpose they
+// serve is to retrieve experiments and can not be used to modify anything
+// within our internal Split.io system.
+var (
+	prod      = "8me2vu6v8lhssdkrpenp1uunl9s3bdc8njqp"
```
ctrombley, line 20: nit: consider something like `prodAPIKey` or `prodKey` for clarity?

internal/split/split_service.go
```diff
@@ -0,0 +1,88 @@
...
+	err = ioutil.WriteFile(filename, blob, 0777)
+	if err != nil {
+		log.Errorf("could not create temp file: %s", err)
+	}
+
+	return filename
+}
+
+func GetAPIKeyByRegion(region string) string {
+	if region == "staging" {
```
ctrombley, line 84: using `strings.EqualFold()` here eliminates the need for case folding elsewhere.

internal/split/split_service.go
```diff
@@ -0,0 +1,88 @@
...
+		return staging
```
ctrombley, line 85: Do we need to support the EU as well? Or is that considered prod for the purpose of split testing?

internal/split/split_service.go
```diff
@@ -0,0 +1,88 @@
...
+)
+
+type Service struct {
+	client *client.SplitClient
+}
+
+// Creates a new instance of a Split Factory
+// Using "localhost" as the apiKey allows us to use Split.io
+// in localhost mode as defined in their documentation
+func NewSplitService(apiKey string, region string) (*Service, error) {
```
ctrombley, line 32: nit: rename to `NewService`

ctrombley: How will users be able to pass API keys if they are not exported from this package?  Also consider if it's worth abstracting the concept of keys from the user and just providing a simple prod/staging switch (bool or enum).  This could allow for initialization with only a single argument rather than the two here.

internal/split/split_service.go
```diff
@@ -0,0 +1,88 @@
...
+	cfg := conf.Default()
+	if apiKey == "localhost" {
+		cfg.SplitFile = CreateMockSplits()
```
ctrombley, line 35: Move assignment to test setup from within a test file?

ctrombley (changes requested):
Nice work here, I have a few items up for consideration.

internal/split/split_service.go
abrunner94: Yes, it's considered part of `prod`

internal/split/split_service.go
abrunner94: Split requires the `apiKey` portion to be `localhost` for unit-testing. I don't see a good way of extracting the `cfg := conf.Default()` portion from the new service initialization, but let me know what you were thinking!

ctrombley: Maybe just export a `SplitFile string` field from the `split` package, and set `cfg.SplitFile = SplitFile` if it is set to anything other than the empty string?  Then we could set that value to the result of `CreateMockSplits()` from inside the test.

ctrombley (approved):

internal/split/split_service.go
Julien4218: Let's move those to GH secret, and also support env var to make our testing easier.

Julien4218 (changes requested):
@ctrombley @abrunner94 
I've tested this by making the suggested change in the readme, using a pretend test feature of "virtuoso_cli_test" and wiring up the snippet below in the recipe_installer.go

	treatment := split.Service.Get(split.VirtuosoCLITest)
	log.Printf("Got treatment: %s for %s", treatment, split.VirtuosoCLITest)

When I run this local on an ec2, I'm seeing an error message. I shouldn't see this ERROR message as a user, also I'd suggest the change the actual text to be a bit less directed to the user (the user certainly didn't wire this split service).

I've added this commit to make it easier to test https://github.com/newrelic/newrelic-cli/commit/14aa78008d01b2bcda2766f5fa669988ff68a4e8

splitioErrorStartup

abrunner94:
@Julien4218 This might be related to the keys being moved to GH secrets. I'm not sure if compiling locally will pull down those secrets.

Julien4218:
> @Julien4218 This might be related to the keys being moved to GH secrets. I'm not sure if compiling locally will pull down those secrets.

That's fair, however my next test was to put some invalid credentials in env vars for those configs. In any ways, we need to make sure that any error related to the split service are handled silently from the user point of view. We can keep those as Debug info.

Julien4218 (approved):
Tested on an EC2 with a local build and no split keys, no error displayed to the user (good), and error logged in newrelic-cli.log file.

PR #991, opened 2021-07-16
fix(install): wire debug log to file

Julien4218:


sanderblue (approved):

ctrombley (dismissed):

go.mod
```diff
@@ -5,7 +5,7 @@ go 1.16
 require (
 	github.com/AlecAivazis/survey/v2 v2.2.14
 	github.com/Masterminds/semver/v3 v3.1.1
-	github.com/briandowns/spinner v1.16.0
+	github.com/briandowns/spinner v1.15.0
```
sanderblue, line 8: Was downgrading this dependency intentional?

internal/install/license_key_fetcher.go
```diff
@@ -34,6 +35,6 @@ func (f *ServiceLicenseKeyFetcher) FetchLicenseKey(ctx context.Context) (string,
 		return "", err
 	}
 
-	log.Debugf("Found license key %s", licenseKey)
+	log.Debugf("Found license key %s", utils.Obfuscate(licenseKey))
```
sanderblue, line 38: 👍

sanderblue (dismissed):

go.mod
Julien4218: Correct @sanderblue we found a regression in the spinner latest (1.16) version, it's duplicating the message when stopping somehow. We reverted to fix the issue.
cc @ctrombley

internal/install/license_key_fetcher.go
Julien4218: @sanderblue for the story, I started implemented an `Obfuscate` method in that same file, and then I thought, this is pretty generic, I should move it to a more global scope, so I search and I found the utils package. That looks like a right place. As I scroll down I found the exact same implementation of the Obfuscate method. I happily deleted my new implementation :)

ctrombley (approved):

ctrombley (approved):

PR #949, opened 2021-07-07
refactor(install): validate infrastructure agent installation directly via the agent's local server

sanderblue:
TODO:

- [ ] Add debug and trace logging where appropriate
- [ ] Put files in proper domains
- [ ] Add tests for agent validator

PR #994, opened 2021-07-19
feat(install): [wip] check if latest version of CLI is installed

sanderblue:
⚠️ **WIP** ⚠️ 

To-do:

- [ ] Figure out the best way to export the CLI `version` [variable](https://github.com/newrelic/newrelic-cli/blob/main/cmd/newrelic/main.go#L34). Might be able to access it via Cobra internals
- [ ] Ensure mini http and design pattern will work with the localhost validation as well #949 (cc: @noahmmcgivern)


sanderblue:
Closing in favor of #1036

PR #1001, opened 2021-07-21
fix(install): lower log level to debug when retry failures occur during validation

ctrombley:


Julien4218 (commented):
@ctrombley this is good.
There is also a quick failure when we install quickly after the account creation. It looks like one of the first web request is failing after a few attempts and not wait/retrying.
See the output in https://newrelic.atlassian.net/secure/RapidBoard.jspa?rapidView=1779&projectKey=VIRTUOSO&modal=detail&selectedIssue=VIRTUOSO-1057&quickFilter=6464

PR #1005, opened 2021-07-22
fix(logging): pass logrus instance to NR client

ctrombley:
Depends on https://github.com/newrelic/newrelic-client-go/pull/780


sanderblue (approved):
Get to the 🚁 !! 

🐑  it now! 😃

PR #1014, opened 2021-07-23
fix: failure events

abrunner94:


ctrombley (approved):

Julien4218 (approved):
Fix works, I can now see a FAILED event with an errorMessage populated

PR #1021, opened 2021-07-28
chore(tutone): update generated code

nr-developer-toolkit:
Update generated code using tutone
- make generate
- PR generated by [create-pull-request][1]

[1]: https://github.com/peter-evans/create-pull-request

### DIFF
--- a/.tutone.yml
+++ b/.tutone.yml
@@ -41,36 +41,6 @@ packages:
           - name: apiAccessDeleteKeys
             clientPackageName: apiaccess
             clientMethod: nrClient.APIAccess.DeleteAPIAccessKey
-  - name: types
-    path: internal/install/types
-    generators:
-      - typegen
-    types:
-      - name: OpenInstallationDocsStitchedFields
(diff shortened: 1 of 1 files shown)


internal/install/types/types.go
```diff
@@ -376,8 +417,6 @@ type OpenInstallationRecipe struct {
 	// NRQL the newrelic-cli uses to validate this recipe
 	// is successfully sending data to New Relic
 	ValidationNRQL NRQL `json:"validationNrql,omitempty"`
-	// validation url to validate with infra health endpoint
-	ValidationURL string `json:"validationUrl,omitempty"`
```
sanderblue, line 380: This was manually added IIRC. We need to add this to the backend for code generation to work via introspection.

sanderblue (changes requested):
Need to update the backend service with the `validationUrl` property.

ctrombley:
It seems to me that we should disable code generation for the `install` command now that we are no longer using a backend service.

PR #1022, opened 2021-07-29
feat(utils): add context to retry helper

ctrombley:


Julien4218 (approved):
@ctrombley this is great, exactly what I was hoping for.
One slight adjustment perhaps, I wonder if we could make the signature of ExecWithRetries( ) simpler by just returning the RetryContext without the last error.
It seems if the caller cares about error, the caller should fetch it directly in the returned RetryContext?
I could eventually see a Canceled bool in the RetryContext to capture that this was a canceled operation, to avoid having the consumer look at the last error to determine that it was. I also don't know we need to append an error when the operation is canceled if we capture the bool directly, just some thoughts.

internal/utils/retry.go
```diff
@@ -37,13 +55,15 @@ func (r *Retry) ExecWithRetries(ctx context.Context) error {
 
 			select {
 			case <-ctx.Done():
-				return context.Canceled
+				retryCtx.Errors = append(retryCtx.Errors, context.Canceled)
```
Julien4218, line 58: @ctrombley should we update retryCtx.Canceled = true when context.Canceled?

internal/diagnose/config_validator.go
```diff
@@ -108,7 +110,13 @@ func (c *ConfigValidator) validateKeys(ctx context.Context) error {
 	}
 
 	r := utils.NewRetry(c.PostMaxRetries, c.PostRetryDelaySec*1000, validateKeyFunc)
-	return r.ExecWithRetries(ctx)
+	retryCtx := r.ExecWithRetries(ctx)
+
+	if !retryCtx.Success {
+		return retryCtx.Errors[len(retryCtx.Errors)-1]
```
Julien4218, line 116: @ctrombley maybe use the new method MostRecentError( ) here too

internal/utils/retry.go
ctrombley: 👍 I missed this, thanks.

PR #1032, opened 2021-08-03
feat(install-events-service): updated schema for reporter

abrunner94:
## Context

This PR is adding new fields to each recipe status as well as a new mutation for creating an install status event within the `InstallComplete` implementation of the interface.


ctrombley (approved):
LGTM

PR #1036, opened 2021-08-04
feat(install): prompt when installed CLI is not on the latest version

sanderblue:
If a user is attempting to run `newrelic install` with an older version of the CLI, we'll now be exiting the install process and prompting the user to install the latest version of the CLI. 

Example output when running `newrelic install` on a version that's behind the latest release version.

Screen Shot 2021-08-18 at 3 40 48 PM

### To-do:

- [x] Centralize where we set and cache the CLI's version string
    - Set in `internal/cli/version.go` so it can be shared across packages
    - Remove old references to old version vars and use `cli.Version` instead of `main.version` (including `LDFLAGS`)
- [x] More tests needed
- [x] Best way to send install status event when this scenario occurs? 
- [x] Ensure Windows support (the update prompt command will be different)


build/compile.mk
```diff
@@ -6,7 +6,7 @@ GO         ?= go
 BUILD_DIR  ?= ./bin
 PROJECT_MODULE ?= $(shell $(GO) list -m)
 # $b replaced by the binary name in the compile loop, -s/w remove debug symbols
-LDFLAGS    ?= "-s -w -X main.version=$(PROJECT_VER) -X main.appName=$$b -X $(PROJECT_MODULE)/internal/client.version=$(PROJECT_VER)"
+LDFLAGS    ?= "-s -w -X main.version=$(PROJECT_VER) -X main.appName=$$b -X $(PROJECT_MODULE)/internal/client.version=$(PROJECT_VER) -X $(PROJECT_MODULE)/internal/cli.Version=$(PROJECT_VER)"
```
sanderblue, line 9: `main.version` and `client.version` should be able to reference `cli.Version` now since we export it from the package

internal/cli/version.go
```diff
@@ -0,0 +1,141 @@
...
+	}
+
+	prereleaseString := v.Prerelease()
+	hasPrereleaseString := prereleaseString != ""
+
+	if hasPrereleaseString && strings.Contains(prereleaseString, "dirty") {
+		return true
+	}
+
+	if hasPrereleaseString && !strings.Contains(prereleaseString, "alpha") && !strings.Contains(prereleaseString, "beta") {
```
sanderblue, line 132: When we're in a dev environment, we're "always behind" the latest version, however, some prerelease strings might contain `alpha` or `beta` (e.g. `v1.0.0-beta` etc). Technically we've never release this kind of tag, but I wanted to have defensive code in place just in case. 

This scenario is fairly unlikely for the foreseeable future, so we can remove it if we don't want it. 

e.g.

internal/install/execution/installevents_reporter.go
```diff
@@ -154,6 +154,16 @@ func (r InstallEventsReporter) createInstallStatusEvent(state installevents.Inst
 	return err
 }
 
+func (r InstallEventsReporter) UpdateRequired(status *InstallStatus) error {
+	err := r.createInstallStatusEvent(
+		installevents.InstallationInstallStateType(installevents.InstallationRecipeStatusTypeTypes.UNSUPPORTED),
```
sanderblue, line 159: This meets the current requirements, but wondering if we need an additional state enum/scalar to better categorize this.

internal/install/recipe_installer.go
```diff
@@ -151,6 +152,33 @@ func (i *RecipeInstaller) Install() error {
 		return err
 	}
 
+	// If not in a dev environemt, check to see if
+	// the installed CLI is up to date.
+	// if !cli.IsDevEnvironment() {
```
sanderblue, line 157: **TODO:** Uncomment when ready to merge

internal/install/execution/install_status.go
```diff
@@ -43,6 +43,7 @@ type InstallStatus struct {
 	InstalledPacks            []*ObservabilityPackStatus `json:"packslInstalled"`
 	RedirectURL               string                     `json:"redirectUrl"`
 	HTTPSProxy                string                     `json:"httpsProxy"`
+	UpdateRequired            bool
```
sanderblue, line 46: ❗ Reminder: get clarity on the [`SetVersions()`](https://github.com/newrelic/newrelic-cli/pull/1036/files#diff-a4104ffcec79e09559961647e745b2de0686e79944e35d2b8a7de7811830c714R387-R394) method

sanderblue: After digging a bit, we can definitely use `cli.Version()` here as well. I've updated accordingly.

internal/cli/version.go
```diff
@@ -0,0 +1,185 @@
...
+	latestVersion string
+)
+
+const installCLISnippetLinux = `curl -Ls https://raw.githubusercontent.com/newrelic/newrelic-cli/master/scripts/install.sh | bash && sudo newrelic install`
+const installCLISnippetWindows = `[Net.ServicePointManager]::SecurityProtocol = 'tls12, tls'; (New-Object System.Net.WebClient).DownloadFile("https://github.com/newrelic/newrelic-cli/releases/latest/download/NewRelicCLIInstaller.msi", "$env:TEMP\NewRelicCLIInstaller.msi"); msiexec.exe /qn /i $env:TEMP\NewRelicCLIInstaller.msi | Out-Null; & 'C:\Program Files\New Relic\New Relic CLI\newrelic.exe' install`
+
+// NewRelicCLILatestReleaseURL is the URL used to fetch the latest release data.
+const NewRelicCLILatestReleaseURL string = "https://api.github.com/repos/newrelic/newrelic-cli/releases/latest"
+
+// NewRelicCLILatestReleaseURL is the URL used to fetch the latest release data.
```
ctrombley, line 32: Nit: This docstring needs an update.

internal/cli/version.go
```diff
@@ -0,0 +1,185 @@
...
+const PrereleaseEnvironmentMsgFormat string = `
```
ctrombley, line 33: Can we just say "It appears you're using prerelease version X?"

internal/cli/version.go
```diff
@@ -0,0 +1,185 @@
...
+
+  To update your CLI and continue this installation, run this command:
+
+    %s
+`
+
+// Version returns the version of the CLI that's currently installed.
+func Version() string {
+	if version == "" {
+		version = os.Getenv("NEW_RELIC_CLI_VERSION")
```
ctrombley, line 54: Why is this tied to an environment variable?  If it's for testing purposes, can't we just update the value of the `version` field from within the tests?

internal/cli/version.go
```diff
@@ -0,0 +1,185 @@
...
+//  v0.32.1-10-gbe63a24-dirty
+//
+// In this example version string, "10" represents the number of commits
+// since the 0.32.1 tag was created. The "gbe63a24" is the previous commit's
+// abbreviated sha. The "dirty" part means that git was in a dirty state at compile
+// time, meaning an updated file was saved, but not yet committed.
+func IsDevEnvironment() bool {
+	v, err := semver.NewVersion(version)
+	if err != nil {
+		return true
```
ctrombley, line 125: Does it make sense to return true here if we have a nonsensical version string?  Would it make more sense to return the error?

internal/cli/version.go
```diff
@@ -0,0 +1,185 @@
...
+	}
+
+	prereleaseString := v.Prerelease()
+	if prereleaseString == "" {
```
ctrombley, line 129: Wouldn't we just want to `return prereleaseString != ""` here and be done with this function?

internal/install/recipe_installer.go
```diff
@@ -155,6 +156,33 @@ func (i *RecipeInstaller) Install() error {
 		return err
 	}
 
+	// If not in a dev environemt, check to see if
```
ctrombley, line 159: Nit: Extract to method?

internal/cli/version.go
sanderblue: Glad you asked. This variable isn't actually used anymore since the `IsDevEnvironment()` function makes sure we don't even ask about updating. I just forgot to delete this from my copy/paste from my previous POC.

internal/cli/version.go
sanderblue: It was due to another test that utilizes an environment variable. Here is [the test](https://github.com/newrelic/newrelic-cli/blob/main/internal/install/recipe_installer_test.go#L904) that uses it, and [here](https://github.com/newrelic/newrelic-cli/blob/c9fe182cc5b8893ba5187f7e968d36aee4b25206/internal/install/execution/install_status.go#L386-L393) is where it was implemented.

internal/cli/version.go
sanderblue: Fair question and open to suggestions. This was a weird decision to make. The only time I've really seen a nonsensical version string is during development, but I guess there's still a chance of us actually releasing a really weird tag, although it's highly unlikely and would likely get rectified soon after.

internal/cli/version.go
sanderblue: Since I can't remember if I reversed this logic on purpose, this seems like a good chance to make.  Good call.

internal/install/recipe_installer.go
sanderblue: Seems reasonable. Will do.

internal/cli/version.go
sanderblue: Did some thinking about this. I remember this being odd with how version strings can be handled. This logic pattern is a bit more defensive. I think we should leave it for now.

Julien4218 (approved):

PR #1069, opened 2021-08-26
feat(install): provide redirect URL for fallback installation, add installation summary

sanderblue:
Provides a redirect URL after an incomplete installation process that allows the user to return to the UI and attempt to finish incomplete installations.

This also provides an installation summary similar to what's displayed in the UI at the end of the installation process.


internal/install/execution/platform_link_generator.go
```diff
@@ -39,6 +45,10 @@ func (g *PlatformLinkGenerator) GenerateEntityLink(entityGUID string) string {
 // also provided in the nerdstorage document. This provides the user two options
 // to see their data - click from the CLI output or from the frontend.
 func (g *PlatformLinkGenerator) GenerateRedirectURL(status InstallStatus) string {
+	if status.HasFailedRecipes || status.HasCanceledRecipes {
+		return g.GenerateExplorerLink("")
```
Julien4218, line 49: @sanderblue do we need to send any filter, hostname in the URL? if not  the URL should be simpler without the `platform[filters]=[hostname]` parts.
Maybe we should have a different method specific to this use case.

Also, wouldn't we want to have a same URL for success? We essentially want the user to go back to the NR1 Installation Plan page regardless of success/failure/cancel?

sanderblue: We need the `platform[filters]` because of the `successLinkConfig` parameter in recipes.

Regarding when we add the referrer param: I was asking myself the same question. I think we can append this param in all situation and let the UI decide what it does with it.

sanderblue: Making some updates. Will re-request review tomorrow.

internal/install/command.go
```diff
@@ -63,7 +63,10 @@ var Command = &cobra.Command{
 				return
 			}
 
-			log.Fatalf("We encountered an error during the installation: %s. If this problem persists please visit the documentation and support page for additional help here: https://one.newrelic.com/-/06vjAeZLKjP", err)
+			fmt.Println()
+			fmt.Printf("We encountered an error during the installation: %s.\n\n", err)
```
Julien4218, line 67: @sanderblue I think that whole phrasing should be reviewed (changed), I dont think we should say "error".
Check with Sam/Aaron, perhaps more something like "We couldn't complete this installation, details :%s" and send the user back to the UI.

internal/install/execution/terminal_reporter.go
```diff
@@ -87,25 +89,70 @@ func (r TerminalStatusReporter) InstallComplete(status *InstallStatus) error {
...
+		suffix := fmt.Sprintf("(%s)", strings.ToLower(string(s.Status)))
+
+		if s.Status == RecipeStatusTypes.INSTALLED {
+			icon = successIcon
+			suffix = fmt.Sprintf("(%s)", color.GreenString(strings.ToLower(string(s.Status))))
+		}
+
+		if s.Status == RecipeStatusTypes.FAILED {
+			icon = failedIcon
+			suffix = fmt.Sprintf("(%s)", color.RedString(strings.ToLower(string(s.Status))))
```
Julien4218, line 137: @sanderblue we should check the color coding, Red might be viewed as too harsh (making users think of an error), maybe yellow, worth checking with Sam/Aaron and/or Daniel

internal/install/command.go
sanderblue: Agreed. I mentioned it to Sam. In general the output needs additional organization to make it friendlier to the users' eye.

internal/install/recipe_installer.go
```diff
@@ -350,8 +350,7 @@ func (i *RecipeInstaller) installRecipes(ctx context.Context, m *types.Discovery
 			lastError = err
 
 			log.Debugf("Failed while executing and validating with progress for recipe name %s, detail:%s", r.Name, err)
```
Julien4218, line 352: @sanderblue another "Failed" message here needing to be updated.

internal/install/command.go
Julien4218: I assume you'll reach out to him to finalize those details? Let me know when done and I'll start testing.

sanderblue: I'll confirm tomorrow.

internal/install/recipe_installer.go
sanderblue: Do we care about that for debug messages vs default user-facing messages?

Julien4218: Ho sorry I didnt see this was debug and not the default output for users, that's fine then.

internal/install/command.go
sanderblue: Closing the loop, we're changing this to a debug log since the general output points the customer back to finishing any incomplete installations. As far as I know, we only have one scenario where we run into this error and that's when the recipes are not compiled, so likely will only occur during local dev. I've accounted for both of these scenario 😎

internal/install/execution/terminal_reporter.go
sanderblue: Updated to yellow and a less harsh icon :)

Julien4218:
@sanderblue I've reached out in the team channel regarding one of the user message than I found misleading.
Also when I run the test local, I see this test failing, any idea why?
=== FAIL: internal/install/execution TestGenerateRedirectURL_InstallSuccess (0.00s)
    platform_link_generator_test.go:41:
                Error Trace:    platform_link_generator_test.go:41
                Error:          Not equal:
                                expected: "https://staging-one.newrelic.com/redirect/entity/ABC123"
                                actual  : "https://one.newrelic.com/redirect/entity/ABC123"

Julien4218:
Test is now fixed, thanks @sanderblue

Julien4218 (changes requested):
@sanderblue, I'm holding this work until we have the green light to ship, since the UI will be behind FF but this work isn't (we discussed in planning not needing to this at the CLI level).

internal/install/execution/platform_link_generator_integration_test.go
```diff
@@ -43,5 +43,15 @@ func TestGenerateShortNewRelicURL_InvalidAPIKey(t *testing.T) {
...
 	require.Error(t, err)
-	require.Equal(t, len(result), len(longURL))
+	require.Equal(t, result, longURL)
+}
+
+func TestGenerateShortNewRelicURL_NonNewRelicURL(t *testing.T) {
+	t.Parallel()
+
+	g := NewPlatformLinkGenerator()
+	longURL := "https://www.youtube.com/watch?v=dQw4w9WgXcQ&ab_channel=RickAstley"
```
ctrombley, line 53: nit: remove `&ab_channel=RickAstley` from the URL so as not to give away the punchline

Julien4218 (approved):

PR #1059, opened 2021-08-24
Feat command to migrate configuration file v3->v4

paologallinaharbur:
### Purpose of the PR

We wrote a small helper in order to simplify the migration between V3 and V4 configuration of onHost integration.

Both the users and the support could be interested in this tool since given a V3 definition and configuration generates a working V4 configuration file.

### Doubts

I was not 100% sure that this is the right place for the new functionality or if it is better to place it under the `agent` command.

PR #1075, opened 2021-08-30
Feat command to migrate configuration file v3->v4

ctrombley:
Continuing work from #1059.


paologallinaharbur:
We were discussing if it is better to place the new subcommand under agent `newrelic agent config migrateIntegrationsV3toV4` instead of creating a completely new command integration `newrelic integrations config migrateV3toV4`

internal/integrations/model.go
```diff
@@ -0,0 +1,72 @@
+package integrations
+
+import "time"
+
+// This code is taken directly from the infrastructure agent
+// To avoid an extra dependency it has been copied/pasted
```
roobre, line 6: Can we link the exact commit where this comes from? That way we should get some sort of traceability.

internal/integrations/utils.go
```diff
@@ -0,0 +1,194 @@
+package integrations
+
+import (
+	"fmt"
+	log "github.com/sirupsen/logrus"
+	"gopkg.in/yaml.v2"
```
roobre, line 6:
```suggestion
	"gopkg.in/yaml.v3"
```

internal/integrations/utils.go
```diff
@@ -0,0 +1,194 @@
...
+
+	// The field os does not have currently a simple way to be migrated
+	if v3Definition.OS != "" {
+		log.Debugf("The old definitions had a os directive, %s. Usually it is not needed, use `when` field otherwhise", v3Definition.OS)
+	}
+
+	v4Config := &v4{}
+	for commandName, pluginV1Command := range v3Definition.Commands {
+		for _, pluginV1Instance := range v3Configuration.Instances {
+			if commandName == pluginV1Instance.Command {
```
roobre, line 86: Should we log an error on `else`?

internal/integrations/utils.go
```diff
@@ -0,0 +1,194 @@
...
+		}
+	}
+}
+
+func writeOutput(pathOutput string, v4Config *v4, definitionBytes []byte, configurationBytes []byte) error {
+	if v4Config == nil {
+		return fmt.Errorf("v4Config pointer is nil")
+	}
+
+	file, err := os.OpenFile(pathOutput, os.O_RDWR|os.O_CREATE|syscall.O_TRUNC, 0666)
```
roobre, line 146: I'd rather check for existence and rename the old file to `.old` or something along those lines. Getting data destroyed by a CLI can be frustrating :)

internal/integrations/utils.go
```diff
@@ -0,0 +1,194 @@
...
+	if err != nil {
+		return fmt.Errorf("adding old configuration as comment, %w", err)
+	}
+
+	return nil
+}
+
+func writeV4Config(v4Config *v4, file *os.File) error {
+	// see https://github.com/go-yaml/yaml/commit/7649d4548cb53a614db133b2a8ac1f31859dda8c
+	yaml.FutureLineWrap()
```
roobre, line 172: Won't be needed with `v3`, according to the link

internal/integrations/utils.go
```diff
@@ -0,0 +1,194 @@
...
+	// Reading old definition file
+	bytes, err := ioutil.ReadFile(path)
+	if err != nil {
+		return nil, fmt.Errorf("reading %s, %w", path, err)
+	}
+
+	err = yaml.Unmarshal(bytes, out)
+	if err != nil {
+		return nil, fmt.Errorf("unmashalling %s, %w", path, err)
+	}
```
roobre, line 68: Could be shorter with `yaml.Decoder`

roobre:
> We were discussing if it is better to place the new subcommand under agent

> `newrelic agent config migrateIntegrationsV3toV4`

Agent subcommand seems slightly better to me, I think.

internal/integrations/utils.go
paologallinaharbur: v2 is already used across the code, I did not want to update a dependency we do not own or import two versions of yaml

internal/integrations/utils.go
paologallinaharbur: v2 is already used across the code, I did not want to update a dependency we do not own or import two versions of yaml

internal/integrations/model.go
paologallinaharbur: I will add it, unfortunately, there is not a single commit for that config, since it has been modified several times in time

internal/integrations/utils.go
```diff
@@ -0,0 +1,194 @@
...
+	yaml.FutureLineWrap()
+	v4ConfigBytes, err := yaml.Marshal(*v4Config)
+	if err != nil {
+		return fmt.Errorf("marshallig v4Config, %w", err)
+	}
+
+	_, err = file.Write(v4ConfigBytes)
+	if err != nil {
+		return fmt.Errorf("writing v4ConfigBytes, %w", err)
+	}
```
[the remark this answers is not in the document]
paologallinaharbur: yes, definetely

internal/integrations/utils.go
paologallinaharbur: yep

internal/integrations/utils.go
paologallinaharbur: mmm no, it is very common that in the definition there are many commands (example: all_data, metrics, inventory) and in the configuration there is only one

internal/integrations/utils.go
paologallinaharbur: I added a overwrite option, and failing in case the file does exist already. I do not like to move around user files (and it would create a loop of .old.old.old 😰 )

roobre (approved):
LGTM, nice work on this!

Julien4218:
@ctrombley what is this PR for?

paologallinaharbur:
Hello, I created this PR  #1059 then it was merged on a local branch to be able to run tests, the context is the following:

```
We wrote a small helper in order to simplify the migration between V3 and V4 configuration of onHost integration.

Both the users and the support could be interested in this tool since given a V3 definition and configuration generates a working V4 configuration file.
```
@Julien4218

Julien4218 (approved):

PR #1099, opened 2021-09-07
feat(install): validate recipes in parallel with all available methods

ctrombley:


Julien4218 (approved):
@ctrombley this looks pretty good to me.

sanderblue (approved):

PR #1101, opened 2021-09-08
feat(install): bypass fastly when fetching recipes

ctrombley:


internal/cli/version.go
```diff
@@ -25,9 +25,8 @@ var (
 const installCLISnippetLinux = `curl -Ls https://download.newrelic.com/install/newrelic-cli/scripts/install.sh | bash && sudo newrelic install`
 const installCLISnippetWindows = `[Net.ServicePointManager]::SecurityProtocol = 'tls12, tls'; (New-Object System.Net.WebClient).DownloadFile(\"https://download.newrelic.com/install/newrelic-cli/scripts/install.ps1\", \"$env:TEMP\\install.ps1\");", "& $env:TEMP\\install.ps1;", "& 'C:\\Program Files\\New Relic\\New Relic CLI\\newrelic.exe' install`
 
-// NewRelicCLILatestReleaseURL is the URL used to fetch the latest release data utilizing GitHub's API.
-// GitHub API Docs: https://docs.github.com/en/rest/reference/repos#get-the-latest-release
-const NewRelicCLILatestReleaseURL string = "https://download.newrelic.com/install/newrelic-cli/currentVersion.txt"
+// NewRelicCLILatestReleaseURL is the URL used to fetch the latest CLI release.
+const NewRelicCLILatestReleaseURL string = "https://s3.us-east-1.amazonaws.com/nr-downloads-main/install/newrelic-cli/currentVersion.txt"
```
Julien4218, line 29: @ctrombley I think we should not use S3 here, because the most up to date version will be returned. During a fresh cli release, the user would have downloaded a potentially out-dated cli (up to 10/15 minutes) and this would be a consistent endpoint until the CDN cache refreshes.
It's ok for the cli version used by the user to be behind the "true" latest version, it's at least consistent between the check of the version and what the user has downloaded. Eventually the cache will refresh and the user will use latest.

build/recipes.mk
```diff
@@ -3,7 +3,7 @@
 # Makefile fragment for recipe actions
 #
 EMBEDDED_RECIPES_PATH         ?= internal/install/recipes/files
-RECIPE_ARCHIVE_VERSION        ?= $(shell curl -Ls https://download.newrelic.com/install/open-install-library/currentVersion.txt)
+RECIPE_ARCHIVE_VERSION        ?= $(shell curl -Ls https://s3.us-east-1.amazonaws.com/nr-downloads-main/install/open-install-library/currentVersion.txt)
```
Julien4218, line 6: out of curiosity, what would happen to dev users building when we move our download to another s3 bucket or region? Would it be obvious that this is the place to update?

Julien4218 (changes requested):
@ctrombley could you undo the change in version.go, I put a comment in there, I don't think we want to have a different URL when validating the version on the CLI.

internal/cli/version.go
ctrombley: 👍

build/recipes.mk
ctrombley: This is the only place to update, correct.

Julien4218: I think so

Julien4218 (approved):

build/recipes.mk
```diff
@@ -3,7 +3,7 @@
...
 RECIPE_ARCHIVE_URL            ?= https://download.newrelic.com/install/open-install-library/${RECIPE_ARCHIVE_VERSION}/recipes.zip
```
Julien4218, line 7: @ctrombley don't we need to also update RECIPE_ARCHIVE_URL

Julien4218 (changes requested):

Julien4218 (approved):

build/recipes.mk
ctrombley: 👍

PR #1110, opened 2021-09-13
fix(install): remove redundant recipe validation progress messages

ctrombley:


Julien4218 (approved):
Tested in linux, looks good.

PR #1137, opened 2021-09-28
feat(install): send DETECTED status when New Relic compatibility is detected

sanderblue:


internal/install/execution/install_status.go
```diff
@@ -117,6 +119,20 @@ func (s *InstallStatus) DiscoveryComplete(dm types.DiscoveryManifest) {
 	}
 }
 
+// RecipeCompatilityDetected handles setting the DETECTED status for the provided
+// recipe as well sending out the DETECTED status event to the install events service.
+// RecipeCompatilityDetected is called when a recipe is available and passes the checks in both
+// the process match and the pre-install steps of recipe execution.
+func (s *InstallStatus) RecipeCompatilityDetected(recipe types.OpenInstallationRecipe) {
```
Julien4218, line 126: I would keep the naming consistent with "RecipeDetected" and drop Compatibility which otherwise is confusing.

internal/install/recipe_installer_test.go
```diff
@@ -892,13 +890,85 @@ func TestInstall_GuidReport(t *testing.T) {
...
 	require.Equal(t, rv.ValidateVal, statusReporters[0].(*execution.MockStatusReporter).RecipeGUID[testRecipeName])
-	require.Equal(t, status.CLIVersion, "v0.0.1")
+	require.Equal(t, status.CLIVersion, cli.Version())
 	require.Equal(t, 3, len(statusReporters[0].(*execution.MockStatusReporter).Durations))
 	for _, duration := range statusReporters[0].(*execution.MockStatusReporter).Durations {
 		require.Less(t, int64(0), duration)
 	}
 }
 
+func TestInstall_DetectedRecipe(t *testing.T) {
```
Julien4218, line 900: Maybe update the name to highlight the intent of the test, for example TestInstallShouldDetectNothing, or Zero

internal/install/recipe_installer_test.go
```diff
@@ -892,13 +890,85 @@ func TestInstall_GuidReport(t *testing.T) {
...
+	v = validation.NewMockRecipeValidator()
+
+	i := RecipeInstaller{ic, d, l, mv, f, e, v, ff, status, p, pi, sp, lkf, cv, rvp, rf, av}
+
+	err := i.Install()
+
+	require.NoError(t, err)
+	require.Equal(t, 2, statusReporters[0].(*execution.MockStatusReporter).RecipeDetectedCallCount)
+}
+
```
Julien4218, line 971: Could we add test with process detection? A recipe with a process detected should raise the Detected status.
A recipe with process detected, and an exit code not 132 should NOT be detected.
A recipe with process detected, and no script should be detected.

internal/install/recipes/recipe_filter_runner.go
```diff
@@ -73,6 +73,7 @@ func (rf *RecipeFilterRunner) RunFilterAll(ctx context.Context, r []types.OpenIn
 		filtered := rf.RunFilter(ctx, &recipe, m)
 
 		if !filtered {
+			rf.installStatus.RecipeCompatilityDetected(recipe)
```
Julien4218, line 76: I'm not sure this implements the logic defined in the story, but one way to validate would be to add the tests that cover the story ACs, I've also put some comment in the recipe_installer_test.go file

Julien4218 (changes requested):
@sanderblue I left a few comments, I think we need to have some tests to validate the AC, and I'm not sure this is working as expected currently.

internal/install/recipe_installer_test.go
sanderblue: Sure thing

internal/install/recipes/recipe_filter_runner.go
sanderblue: I'll double check, but after our conversation I think this is where we wanted it. I'll add the test mentioned above to verify this is where we want this called.

internal/install/recipe_installer_test.go
sanderblue: Added

internal/install/recipe_installer_test.go
```diff
@@ -892,13 +890,116 @@ func TestInstall_GuidReport(t *testing.T) {
...
+func TestInstall_ShouldNotDetectRecipeWithProcessMatch(t *testing.T) {
```
Julien4218, line 956: I think this test is misnamed, it is validating that we detect when process match and scripts indicate the specific value of 132.

Julien4218: We also need the flip side of this test, a test that validate we don't detect when process match, and the script returns an exit code that is NOT 132.

Julien4218 (changes requested):

internal/install/recipe_installer_test.go
sanderblue: This test is difficult to name. It's testing that a process match was found, but the preinstall check failed with exit code 132, meaning we did _not_ detect compatibility. But yes we need the flip side of this too.

Julien4218: @sanderblue actually this test is incorrect, check back the AC in the story. If process matches, and script is present with an exit code of 132, the Detected event should be fired up.

internal/install/recipe_installer_test.go
```diff
@@ -892,17 +890,166 @@ func TestInstall_GuidReport(t *testing.T) {
...
+
+	i := RecipeInstaller{ic, d, l, mv, f, e, v, ff, status, p, pi, sp, lkf, cv, rvp, rf, av}
+
+	err := i.Install()
+
+	require.NoError(t, err)
+	require.Equal(t, 2, statusReporters[0].(*execution.MockStatusReporter).RecipeDetectedCallCount)
+}
+
+func TestInstall_ShouldNotDetectRecipes(t *testing.T) {
```
Julien4218, line 956: @sanderblue the test name is incorrect, since the test is asserting we're finding 1 Detected.
I would break this test into multiple ones:
- TestInstall_ShouldDetect_PreInstallDetected
- TestInstall_ShouldDetect_PreInstallOk
- TestInstall_ShouldNotDetect_PreInstallError

internal/install/recipe_installer_test.go
```diff
@@ -892,17 +890,166 @@ func TestInstall_GuidReport(t *testing.T) {
...
 	require.Equal(t, rv.ValidateVal, statusReporters[0].(*execution.MockStatusReporter).RecipeGUID[testRecipeName])
-	require.Equal(t, status.CLIVersion, "v0.0.1")
+	require.Equal(t, status.CLIVersion, cli.Version())
 	require.Equal(t, 3, len(statusReporters[0].(*execution.MockStatusReporter).Durations))
 	for _, duration := range statusReporters[0].(*execution.MockStatusReporter).Durations {
 		require.Less(t, int64(0), duration)
 	}
 }
 
+func TestInstall_ShouldDetectRecipe(t *testing.T) {
```
Julien4218, line 900: This is a good test, this seems to detect when there is no process detection and no script.

PR #1142, opened 2021-10-05
fix: platform detection for opensuse-leap

noahmmcgivern:


Julien4218 (changes requested):
Actually if there are new values for Platform, those should be added to nerdgraph first, and then re-generate the type for that enum.

PR #1165, opened 2021-11-02
chore(release): use GitHub Actions UI to trigger release workflow via `workflow_dispatch` event 

sanderblue:
🤖 **New releases will be created via the GitHub Actions UI** 🤖 

manually-trigger-release-workflow

**Checklist**
- [x] Cutting a new release is triggered from the GitHub Actions UI, which triggers the `workflow_dispatch` event
- [x] Release job is dependent on all tests and workflows passing (especially the end to end tests)
- [x] New tag version is automatically bumped based semantic commits using [**svu**](https://github.com/caarlos0/svu)
- [x] CHANGELOG is properly added to the release with the latest release notes
- [x] Document release flow
- [x] Ensure latest open-install-library is pulled

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,9 +1,18 @@
 name: Release
 
+permissions: write-all
+
 on:
-  push:
-    tags:
-      - 'v*'
+  workflow_dispatch:
+    inputs:
+      version:
+        description: New release version
+        # NOTE...
+        #
+        # TBD: If we make this required, we won't need to do
+        # the automated version bump with svu and gobump
+        required: false
+        default: x.x.x
 
 jobs:
   release:
@@ -42,7 +51,13 @@ jobs:
           SNAPCRAFT_TOKEN: ${{ secrets.SNAPCRAFT_TOKEN }}
           SPLIT_PROD_KEY: ${{ secrets.SPLIT_PROD_KEY }}
           SPLIT_STAGING_KEY: ${{ secrets.SPLIT_STAGING_KEY }}
-        run: make release-publish
+        run: |
+          git config --global user.name nr-developer-toolkit
+          git config --global user.email 62031461+nr-developer-toolkit@users.noreply.github.com
+
+          ./scripts/release.sh
+
+          make release-publish
 
       - id: get-version
         uses: battila7/get-version-action@v2.3.0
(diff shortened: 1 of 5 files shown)


scripts/release.sh
```diff
@@ -2,32 +2,56 @@
...
+CHANGELOG_FILE=CHANGELOG.md
+SPELL_CMD=${GOBIN}/misspell
+
+# Compare versions
+VER_CURR=$(${VER_CMD} current --strip-prefix)
+VER_NEXT=$(${VER_CMD} next --strip-prefix)
+
+echo " "
+echo "Comparing tag versions..."
+echo "Current version: ${VER_CURR}"
```
Julien4218, line 32: @sanderblue do you have an example of what VER_CURR would look like? I believe we want something like vX.Y.Z (notice the v prefix).

.github/workflows/release.yml
```diff
@@ -36,13 +50,19 @@ jobs:
...
-          GITHUB_TOKEN: ${{ secrets.DEV_TOOLKIT_TOKEN }}
+          GITHUB_TOKEN: ${{ secrets.RELEASE_TOKEN }}
           DOCKER_USERNAME: ${{ secrets.DOCKER_USERNAME }}
           DOCKER_PASSWORD: ${{ secrets.DOCKER_PASSWORD }}
           SNAPCRAFT_TOKEN: ${{ secrets.SNAPCRAFT_TOKEN }}
           SPLIT_PROD_KEY: ${{ secrets.SPLIT_PROD_KEY }}
           SPLIT_STAGING_KEY: ${{ secrets.SPLIT_STAGING_KEY }}
-        run: make release-publish
+        run: |
+          git config --global user.name nr-developer-toolkit
```
Julien4218, line 60: @sanderblue let's move this user name and email to GH secrets, make sure to document the value in our secret repository.

scripts/release.sh
sanderblue: Example value will be: `0.0.1` 

It will be without the `v` prefix because the variable is set using `--strip-prefix` 

i.e. `svu current --strip-prefix`

We actually may not need to strip the prefix here though. We needed to in the client because we were injecting the value into the client's `version.go` file in the release script, but that's not how the CLI handles this due it being a binary vs a library.

sanderblue: Yep, that was the reason. I can update it so we don't strip the prefix in this script.

sanderblue: Updated

jbeveland27 (approved):
🚢 🇮🇹

Julien4218 (approved):

PR #1166, opened 2021-11-03
feat(nerdgraph query): added support for file inputs

### LINKED ISSUE #1164: Newrelic nerdgraph file argument support
Hello friends.

### Feature Description
If you try to provide long input to `newrelic nerdgraph query` via stdin it's hitting various kernel limits based on the OS.

Example command arguments;
newrelic nerdgraph query "cat mygraph.gql" --variables "cat myDashboard.json"

Example error;
/usr/local/bin/newrelic: Argument list too long

### Describe Alternatives
As an alternative to the stdin option that'd be great if we can also provide a file instead of stdin.

kidk:
Fixes #1164 

I wasn't sure how to test this, so any guidance would be great.


internal/nerdgraph/command_query.go
```diff
@@ -44,13 +52,39 @@ keys are the variables to be referenced in the GraphQL query.
...
+				log.Fatal(err)
+			}
+		}
 
-		err := json.Unmarshal([]byte(variables), &variablesParsed)
+		// parse variables passed as arguments, this will overwrite any existing variables from file
+		err = json.Unmarshal([]byte(variables), &variablesParsed)
 		if err != nil {
 			log.Fatal(err)
 		}
```
jbeveland27, line 74: One small recommendation: rather than allowing `variables` to override `variableFile`, it might be better UX to only allow one or the other so the behavior is more explicit. Then just have an `if/else if` to unmarshal the right variable data into `variablesParsed`.

The `Args` block above could have an explicit check as well:
```
if argsCount == 1 && strings.TrimSpace(variablesFile) != "" && strings.TrimSpace(variables) != "" {
	return errors.New("command expects either variablesFile or variables, not both")
}
```

That's all a minor nit though, and your implementation works as I'd expect. So if you'd rather keep the behavior you have that's cool too.

jbeveland27:
I spoke with @sanderblue and we don't have a good way to test running of commands right now (short of doing some refactoring). So I think what you have is good overall once we decide what to do with the point I raised above ☝🏼

internal/nerdgraph/command_query.go
kidk: I was thinking about the following use case "I have a set of defaults (file), but sometimes I want to overwrite one variable when running" For example staging environments, or specific dashboards names. 

So it is on purpose, but let me check with person who opened the ticket.

kidk: Let's keep it as is @jbeveland27 I hope it's clear from descriptions what's happening. So this is good to merge 🥳

jbeveland27 (approved):

PR #1176, opened 2021-11-05
feat(install): allow recipe target to define a regular expression for matching

Julien4218:
To support specific linux arm versions for the log recipe https://github.com/newrelic/open-install-library/pull/570

### DIFF
--- a/internal/install/recipes/recipe_repository.go
+++ b/internal/install/recipes/recipe_repository.go
@@ -2,6 +2,7 @@ package recipes
 
 import (
 	"math"
+	"regexp"
 	"sort"
 	"strings"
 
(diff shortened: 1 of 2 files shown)


internal/install/recipes/recipe_repository.go
```diff
@@ -157,6 +158,11 @@ func mathMax(numbers []int) int {
 
 func matchRecipeCriteria(hostMap map[string]string, rkey string, rvalue string) bool {
 	if val, ok := hostMap[rkey]; ok {
+		if len(rvalue) > 0 && rvalue[0] == '(' {
```
jbeveland27, line 161: One thing throwing me off here - requiring the regex to be surrounded in parenthesis seems like an opinionated implementation choice that is hard to surface to users (and parentheses actually mean something w.r.t. the captured group in a regex). I've typically seen the start and end of a regex defined by forward slashes (like `/<my_regex>/`). In fact we had the thought to do this early on with our process match regex, but eventually elected to leave off the forward slashes (made it easier, so the regex could just be directly fed into `regexp.Compile`).

The other thought I have - is it even necessary to make a distinction between using a regex vs a literal value? I think specifying a literal version would work for a regex match (i.e. `platformVersion: "10.11"` would match a discovered platform version `10.11`).

jbeveland27 (approved):
Looks good! 🚢

jbeveland27 (approved):

PR #1178, opened 2021-11-09
chore(release): enable forcing a patch release to release open-install-library updates 

sanderblue:


Julien4218 (commented):
Could we do something different where we control this logic in the GHA? 
For example, it seems that if we detect a (feat) then we inc minor, otherwise we increase the patch number all the time.
No need to prompt the user for an input
By the way, here is an example from GH in regard of releasing and handling version https://github.com/actions-ecosystem/action-get-latest-tag/blob/main/.github/workflows/release.yml

sanderblue:
@Julien4218 I looked at that action's workflow earlier as well. The release logic change what you were requesting was trivial and didn't involve needing to change our implementation of that the `actions-ecosystem/action-get-latest-tag` GH action.

Julien4218 (approved):
looking good @sanderblue

PR #1183, opened 2021-11-15
fix(install): set `targetedInstall` to `true` on status when recipe(s) provided in install command

sanderblue:


Julien4218 (approved):
Nice fix @sanderblue

abrunner94 (approved):

PR #1186, opened 2021-11-17
feat(install): add logging url to installation summary

arenjae:
image

### DIFF
--- a/internal/install/execution/platform_link_generator_test.go
+++ b/internal/install/execution/platform_link_generator_test.go
@@ -82,7 +82,7 @@ func TestGenerateLoggingURL_InstallSuccess(t *testing.T) {
 		Statuses:        []*RecipeStatus{agentInstalledStatus, logsInstalledStatus},
 	}
 
-	launcherEncodedParams := "eyJxdWVyeSI6IlwiZW50aXR5Lmd1aWRcIjpcIk1YeEJVRTE4UVZCUVRFbERRVlJKVDA1OE9URTJOelF4TmdcIiJ9"
+	launcherEncodedParams := "eyJxdWVyeSI6IlwiZW50aXR5Lmd1aWQuSU5GUkFcIjpcIk1YeEJVRTE4UVZCUVRFbERRVlJKVDA1OE9URTJOelF4TmdcIiJ9"
 	expectedRedirectURL := fmt.Sprintf("https://%s/redirect/entity/%s", nrPlatformHostname(), infraEntityGUID)
 	expectedLoggingLink := fmt.Sprintf("https://%s/launcher/logger.log-launcher?platform[accountId]=0&launcher=%s", nrPlatformHostname(), launcherEncodedParams)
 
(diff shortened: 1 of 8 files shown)


internal/install/execution/platform_link_generator_test.go
```diff
@@ -45,6 +45,53 @@ func TestGenerateRedirectURL_InstallSuccess(t *testing.T) {
...
+	installStatus := InstallStatus{
+		recipesSelected: []types.OpenInstallationRecipe{infraRecipe, logsRecipe},
+		Installed:       []*RecipeStatus{agentInstalledStatus, logsInstalledStatus},
+		EntityGUIDs:     []string{infraEntityGUID},
+		Statuses:        []*RecipeStatus{agentInstalledStatus, logsInstalledStatus},
+	}
+
+	launcherEncodedParams := "eyJxdWVyeSI6IlwiZW50aXR5Lmd1aWQuSU5GUkFcIjpcIk1YeEJVRTE4UVZCUVRFbERRVlJKVDA1OE9URTJOelF4TmdcIiJ9"
+	expectedRedirectURL := fmt.Sprintf("https://%s/redirect/entity/%s", nrPlatformHostname(), infraEntityGUID)
+	expectedLoggingLink := fmt.Sprintf("https://%s/launcher/logger.log-launcher?platform[accountId]=0&launcher=%s", nrPlatformHostname(), launcherEncodedParams)
```
sanderblue, line 87: 🤔 Solid test, however, I think we need the `[accountId]=0` to be dynamic to match the behavior in the tested function ([reference](https://github.com/newrelic/newrelic-cli/pull/1186/files#diff-f8343147dd77d4baf043d9670284714fdad6657bb697f28765d9400b5536e104R125)). This test fails if I run it locally because I have the `NEW_RELIC_ACCOUNT_ID` environment variable set...which makes wonder how this passed in CI 😆

Another option would be to narrow the scope of assertion to only check that `loggingLinkResult` contains the `launcherEncodedParams` (or some other substring/combo).

internal/install/execution/terminal_reporter.go
```diff
@@ -155,6 +157,24 @@ func (r TerminalStatusReporter) UpdateRequired(status *InstallStatus) error {
 	return nil
 }
 
+func (r TerminalStatusReporter) printLoggingLink(status *InstallStatus) {
+	linkToLogging := ""
+	loggingMsg := "View your logs at the link below:\n"
+	statusesToDisplay := r.getRecipesStatusesForInstallationSummary(status)
+
+	for _, s := range statusesToDisplay {
+		if s.Status == RecipeStatusTypes.INSTALLED && s.DisplayName == "Logs integration" {
```
sanderblue, line 166: We have an exported variable `LoggingRecipeName` that can we used in this conditional, but would refer to `s.Name` 

e.g. `if s.Status == RecipeStatusTypes.INSTALLED && s.Name == types.LoggingRecipeName { // ... }`

internal/install/interfaces.go
```diff
@@ -11,7 +11,7 @@ type ConfigValidator interface {
 	Validate(ctx context.Context) error
 }
 
-// Discoverer is responsible for discovering informataion about the host system.
+// Discoverer is responsible for discovering information about the host system.
```
sanderblue, line 14: 🎉

sanderblue (approved):

sanderblue:
🎉 Thanks for the awesome contribution, @arenjae and @olsonso! 🎉

PR #1188, opened 2021-11-22
feat(install): Add Logging referrer param

olsonso:
Adding the referrer parameter so we can send an event in the logging ui for this :)


sanderblue (approved):
Nice! 🎉

PR #1200, opened 2021-12-15
feat(install): Require powershell to be run as admin

natecanfield822:
This adds the requires tag for admin for powershell.

### DIFF
--- a/scripts/install.ps1
+++ b/scripts/install.ps1
@@ -1,6 +1,6 @@
 $p = New-Object System.Security.Principal.WindowsPrincipal([System.Security.Principal.WindowsIdentity]::GetCurrent())
 if (!$p.IsInRole([System.Security.Principal.WindowsBuiltInRole]::Administrator)) {
-    throw 'This script requires admin privledges to run and the current Windows PowerShell session is not running as Administrator. Start Windows PowerShell by  using the Run as Administrator option, and then try running the script again.'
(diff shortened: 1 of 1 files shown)


natecanfield822:
If you're not logged in as an admin and you don't run the script as an admin, the user will get the below error.

image

Julien4218:
@natecanfield822 this is good, but perhaps not enough.
If a newrelic-cli was previously installed, our copy/paste continue to runs and execute the cli, here is an illustration
image

Julien4218 (changes requested):
Good but not enough

natecanfield822:
Unfortunately it looks like that requires syntax won't stop further command execution, only prevent the script itself from doing anything at all.

This should cause command execution to stop at the script if not run as an admin

scripts/install.ps1
```diff
@@ -1,3 +1,7 @@
+$p = New-Object System.Security.Principal.WindowsPrincipal([System.Security.Principal.WindowsIdentity]::GetCurrent())
+if (!$p.IsInRole([System.Security.Principal.WindowsBuiltInRole]::Administrator)) {
+    throw 'This script requires admin privledges to run and the current Windows PowerShell session is not running as Administrator. Start Windows PowerShell by  using the Run as Administrator option, and then try running the script again.'
```
Julien4218, line 3: @natecanfield822 this seems to work well, however could you double check the wording, for example privileges is misspell. There are also more than 1 spaces between words sometimes.

Julien4218:
It works great, thanks @natecanfield822 
image

Julien4218 (approved):

PR #1204, opened 2022-01-27
chore: adjust pre-commit hook,  fix CI job dependencies

sanderblue:


.github/workflows/release.yml
```diff
@@ -39,6 +39,7 @@ jobs:
 
   release:
     if: github.ref == 'refs/heads/main'
+    needs: [test, e2e, e2e-eu]
```
Julien4218, line 42: @sanderblue 
The release doesn't depend on those to complete first, they can be done in parallel.

Julien4218 (changes requested):
I don't see why we should complete all the tests before starting the release step, this is going to add about 9 min to the overall job time.

sanderblue:
So we should allow a release even if our tests fail? Sorry, but that seems fundamentally wrong.

.github/workflows/release.yml
sanderblue: The release definitely depends on tests passing. That's how we prevent releasing most of our bugs.

sanderblue: I removed the `e2e` and `e2e-eu` dependency from here to speed things up, but we need to keep `test` because we shouldn't release if `test` fails.

sanderblue:
As a compromise, I removed the `e2e` and `e2e-eu` dependency from the `release` job to speed things up, but we need to keep `test` because we shouldn't release if `test` fails. 

I also made it so `release` and `release-windows-installer` run in parallel

Julien4218 (approved):
Yes @sanderblue this is good for now.
We might want to do some more work later, to split the building from the publishing (to github) currently in the "release" step. This way we could parallelize as much until the last step when we could publish to the various locations once we know all the building and tests passed.

Julien4218 (approved):

PR #1205, opened 2022-01-27
Feat/capture metric data in events 2

sanderblue:
⚠️ **WIP** ⚠️ 

Checklist:
- [x] Ensure this works as expected with **targeted** install
- [ ] Ensure this works as expected with **guided** install
- [ ] Update `go_task_recipe_executor.go`
- [x] Update `sh_recipe_executor.go`
- [ ] Update relevant existing tests
- [ ] Add new tests 
- [ ] Document standardized schema for returned JSON from recipe
- [ ] Remove `EnsureDoesNotFilter()` method if no longer used
- [ ] BONUS: Update `SkipRecipeFilterer` to use `CheckCompatibility` method 
- [x] Update install-events-service to accept additional event data as a `map`
- [x] Update newrelic-client-go with new field + types


Julien4218 (dismissed):
@sanderblue approving. Overall it looks good, some name/rename are tricky, but it always is.
It would be great to do a demo before shipping.

Julien4218 (approved):

Julien4218 (changes requested):
We'll redo to support capturing metadata during installation and preInstall successful detection

PR #1211, opened 2022-02-15
chore: replace insights key with license key

noahmmcgivern:


internal/client/client.go
```diff
@@ -108,29 +108,3 @@ func execLicenseKeyRequest(ctx context.Context, client *newrelic.NewRelic, accou
 
 	return "", types.ErrorFetchingLicenseKey
 }
-
-func FetchInsightsInsertKey(accountID int, profileName string) (string, error) {
```
sanderblue, line 112: 🙌 awesome

sanderblue (approved):

PR #1257, opened 2022-03-15
chore(install): support exit code 131(unsupported) during pre-install

NRhzhao:


Julien4218 (approved):
@NRhzhao that looks good to me, I like the simplicity of it.
I think the downside is that we'll surface that in the NR1 UI, but I think it's actually good, so that we (customer and us) start knowing something is not supported.

Julien4218 (approved):

PR #1260, opened 2022-03-15
Dep fix on not available

NRhzhao:


Julien4218 (approved):

internal/install/recipes/bundler.go
```diff
@@ -79,45 +79,32 @@ func (b *Bundler) createBundle(recipes []*types.OpenInstallationRecipe, bType Bu
 
 	for _, r := range recipes {
 		// recipe shouldn't have itself as dependency
-		visited := map[string]bool{r.Name: true}
-		bundleRecipe := b.getBundleRecipeWithDependencies(r, visited)
+		bundleRecipe := b.getBundleRecipeWithDependencies(r)
```
Julien4218, line 82: @NRhzhao we might want to test to make sure a recipe doesn't get multiple duplicates status. Because of the range recipes, I wonder if a recipe would accumulate multiple AVAILABLE status for example if it was in the range loop and as a dependency on another recipe.

Julien4218: Nevermind, I can see the recipe bundler has a check to ensure we dont accumulate duplicates.

Julien4218 (approved):

PR #1268, opened 2022-03-25
chore(install): Update install summary data link

NRhzhao:
* Update so only a Single install would generate a entity link,
  otherwise alwasys show status summary (fix for k8s)
* Remove select feature from install status, it's no longer
  Need after last CLI flow changes
* Update tests

### DIFF
--- a/internal/install/execution/platform_link_generator.go
+++ b/internal/install/execution/platform_link_generator.go
@@ -52,7 +52,12 @@ func (g *PlatformLinkGenerator) GenerateLoggingLink(entityGUID string) string {
 // also provided in the nerdstorage document. This provides the user two options
 // to see their data - click from the CLI output or from the frontend.
 func (g *PlatformLinkGenerator) GenerateRedirectURL(status InstallStatus) string {
-	if status.AllSelectedRecipesInstalled() {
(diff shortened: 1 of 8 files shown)


NRhzhao:
- Update to show spinner when validating install, this is mainly waiting entity ID to be generated

internal/install/execution/platform_link_generator.go
```diff
@@ -52,7 +52,11 @@ func (g *PlatformLinkGenerator) GenerateLoggingLink(entityGUID string) string {
 // also provided in the nerdstorage document. This provides the user two options
 // to see their data - click from the CLI output or from the frontend.
 func (g *PlatformLinkGenerator) GenerateRedirectURL(status InstallStatus) string {
-	if status.AllSelectedRecipesInstalled() {
+	installedCount := len(status.Installed)
+	haveFailure := status.HasUnsupportedRecipes && status.HasFailedRecipes
```
sanderblue, line 56: Just double checking, should this use the `||` operator?

NRhzhao: Thanks, Sander, that was a good catch.  I had the wrong logic on the test builder, hence the test was passing.  Should be good now.

PR #1269, opened 2022-03-25
fix(release): digitally sign install.ps1 script to fix windows installations

sanderblue:
https://newrelic.atlassian.net/browse/VIRTUOSO-2011

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -186,10 +186,8 @@ jobs:
           GITHUB_TOKEN: ${{ secrets.RELEASE_TOKEN }}
         run: echo "::set-output name=upload_url::$(./scripts/get_latest_release_upload_url.sh)"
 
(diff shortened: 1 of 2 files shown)


Julien4218 (changes requested):
@sanderblue this looks pretty good, however I think we need to either do the signing earlier, or move the cp to S3 later, currently it is been copied over on line 142 before we sign.

Julien4218 (changes requested):
@sanderblue I don't believe this is going to work because the SignPS1.cmd is a windows script, but the release job is currently running on linux, https://github.com/newrelic/newrelic-cli/blob/98a8ac24da1af36269ed165a804a8bdc43724811/.github/workflows/release.yml#L45
You'll want to move the sign onto the windows job.

Julien4218 (approved):
This looks good.

PR #1279, opened 2022-04-11
feat(install): handle data usage exceeded scenario

sanderblue:
Depends on [newrelic-client-go #866](https://github.com/newrelic/newrelic-client-go/pull/866)

### DIFF
--- a/internal/diagnose/config_validator.go
+++ b/internal/diagnose/config_validator.go
@@ -14,6 +14,7 @@ import (
 	"github.com/newrelic/newrelic-cli/internal/utils"
 	"github.com/newrelic/newrelic-cli/internal/utils/validation"
 	"github.com/newrelic/newrelic-client-go/newrelic"
+	"github.com/newrelic/newrelic-client-go/pkg/apiaccess"
 	nrErrors "github.com/newrelic/newrelic-client-go/pkg/errors"
 )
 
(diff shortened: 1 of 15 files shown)


internal/utils/retry.go
```diff
@@ -39,6 +41,12 @@ func (r *Retry) ExecWithRetries(ctx context.Context) *RetryContext {
 	for !retryCtx.Success {
 		retryCtx.RetryCount++
 		if err := r.RetryFunc(); err != nil {
+			if _, ok := err.(*nrErrors.PaymentRequiredError); ok {
```
Julien4218, line 44: @sanderblue I would expect a test to validate we don't retry for paywall errors.

internal/install/recipe_installer.go
```diff
@@ -169,20 +170,23 @@ Welcome to New Relic. Let's set up full stack observability for your environment
...
 	errChan := make(chan error)
 	var err error
 
 	err = i.connectToPlatform()
-
 	if err != nil {
-		return err
-	}
+		if _, ok := err.(*nrErrors.PaymentRequiredError); ok {
+			i.status.InstallComplete(err)
```
Julien4218, line 186: It doesn't seem right to have an install complete event without a started event before that. I understand why you've added the complete here though (to get the metric on how many users are having this paywall error), but perhaps the start event should go first.
What happen if there are other NR errors, such as network issues? Perhaps in those cases we shouldn't attempt to write status events?

internal/install/command.go
```diff
@@ -27,7 +28,7 @@ var Command = &cobra.Command{
 	Use:    "install",
 	Short:  "Install New Relic.",
 	PreRun: client.RequireClient,
-	Run: func(cmd *cobra.Command, args []string) {
+	RunE: func(cmd *cobra.Command, args []string) error {
```
Julien4218, line 31: I dont follow why the error handling is changing here. Is this required to surface an error message to the user when we notice the paywal?

internal/diagnose/errors.go
```diff
@@ -10,3 +10,8 @@ var (
 	ErrPostEvent              = errors.New("there was a failure posting data to New Relic. Please try again later or contact New Relic support. For real-time platform status info visit https://status.newrelic.com/")
 	ErrLicenseKey             = errors.New("the configured license key is invalid for the configured account. Please set a valid license key with the `newrelic profile` command. For more details visit https://docs.newrelic.com/docs/apis/intro-apis/new-relic-api-keys/#ingest-license-key")
 )
+
+const PaymentRequiredExceptionMessage = `
+  Your account has exceeded your plan's data limit.
```
Julien4218, line 15: I might remove one of the "Your/your" to lighten up the message a bit, maybe the 2nd one should be a "the".

Julien4218 (commented):
@sanderblue a few feedback items.

internal/install/recipe_installer.go
sanderblue: Good point. I think I had the `InstallStarted` call before this at one point. I'll take a look again and see what I can come up with.

internal/install/command.go
sanderblue: We can also handle this at the subcommand level if that is preferred, but this felt like a top-level type of error scenario. Either way we decide is fine with me.

internal/install/recipe_installer.go
sanderblue: Yep, I had moved this up at one point. I just moved it to where I had it before (just before call `status.SetTargetedInstall()`)

internal/utils/retry.go
sanderblue: Added

internal/install/recipe_installer.go
```diff
@@ -169,24 +170,27 @@ Welcome to New Relic. Let's set up full stack observability for your environment
 		"RecipeNamesProvided": i.RecipeNamesProvided(),
 	}).Debug("context summary")
 
+	i.status.InstallStarted()
+
+	if i.RecipeNamesProvided() {
```
Julien4218, line 175: I believe this SetTargetedInstall( ) should be done before the InstallStarted( ) otherwise the Started event is going to miss the targeted info.

internal/diagnose/config_validator.go
```diff
@@ -71,6 +72,11 @@ func (c *ConfigValidator) Validate(ctx context.Context) error {
 	postEvent := func() error {
 		if err = c.client.Events.CreateEventWithContext(ctx, accountID, evt); err != nil {
 			log.Debug(err)
+
+			if e, ok := err.(*nrErrors.PaymentRequiredError); ok {
```
Julien4218, line 76: I see we don't have any test for the config validator (debt), I'm a bit sorry to ask since the addition here is small, but this feels like it should be unit tested too.
In addition, it took me a bit of time to understand the value of this addition here, versus just returning the error below.

Julien4218 (commented):
@sanderblue I found an issue with the start event missing the targeted info. Also if you could add a test for the config validator that would be great.

internal/install/recipe_installer.go
sanderblue: Didn't realize that, will update.

Julien4218 (approved):
Verified with accountID 3479838

PR #1289, opened 2022-04-29
chore(ci): remove tutone generate workflow

sanderblue:
The `generate.yml` currently provides no value. Removing until this tooling can be integrated more smoothly.


natecanfield822 (approved):
Looks good.

PR #1291, opened 2022-05-02
Nr 1350/detection during target install

NRhzhao:


internal/install/bundle_installer.go
```diff
@@ -105,9 +105,11 @@ func (bi *BundleInstaller) InstallContinueOnError(bundle *recipes.Bundle, assume
 
 func (bi *BundleInstaller) reportBundleStatus(bundle *recipes.Bundle) {
 	for _, recipe := range bundle.BundleRecipes {
-		for _, ds := range recipe.DetectedStatuses {
-			e := execution.RecipeStatusEvent{Recipe: *recipe.Recipe, ValidationDurationMs: ds.DurationMs}
-			bi.statusReporter.ReportStatus(ds.Status, e)
+		if !bi.IsRecipeInstalled(recipe.Recipe.Name) {
```
Julien4218, line 108: @NRhzhao this change seems independent of this work. Is this a needed bug fix? What's the use case illustrating that we don't want to fire detected if the recipe has been installed?

NRhzhao: @Julien4218  Hi Julien, since we doing detection for all recipes during additional bundling.  If we re-report installed recipes status (infra, log), it reset the NR1 UI.
The alternative is to filter out the installed recipes during bundling by passing in installed recipes.  
Let me know if you would prefer doing that way instead.

Julien4218: I see @NRhzhao 
Given we want to detect all the time now (guided+targeted) with the exception of the `-c` mode, would it be easier if we were to move all the detection out of the bundling (before any), perhaps right after we do all the validation (NR connect OK)?

internal/install/recipes/recipe_detector.go
```diff
@@ -10,52 +10,94 @@ import (
...
 
-	if bundleRecipe.AreAllDependenciesAvailable() {
-		status, durationMs := dt.detectRecipe(ctx, bundleRecipe.Recipe)
-		bundleRecipe.AddDetectionStatus(status, durationMs)
-	} else {
-		log.Debugf("Dependency not available for recipe %v", bundleRecipe.Recipe.Name)
+func (dt *RecipeDetector) GetUnavaliableRecipes() (map[string]*RecipeDetectionResult, error) {
+	err := dt.detectRecipes()
+	if err != nil {
+		return nil, err
```
mbruzina, line 82: @NRhzhao: Seems like these are both called only once in `recipe_installer` - maybe this could be `GetDetectedRecipes` and returns two maps - available and unavailable - since these are set in `detectRecipes`.  Then we could get rid of the `isDetectionRunned` concept

NRhzhao: @mbruzina Yeah, I was thinking about that, but opt to went with separate calls since it's simpler.  But I can update to your recommendation.

internal/install/recipes/bundler.go
```diff
@@ -35,7 +34,7 @@ func (b *Bundler) CreateCoreBundle() *Bundle {
 func (b *Bundler) CreateAdditionalGuidedBundle() *Bundle {
 	var recipes []string
 
-	for _, d := range b.Detections {
+	for _, d := range b.AvailableRecipes {
```
mbruzina, line 37: Random thought - instead of iterating over N recipes to remove the two that are part of the core bundle, why not remove the two core bundle recipes from AvailableRecipes since you can delete from a map by key?

NRhzhao: @mbruzina But we still need to loop through our map to create a slice?

mbruzina: Ah, fair point 👍

internal/install/recipes/recipe_detector.go
```diff
@@ -43,16 +42,17 @@ func NewRecipeDetector(contex context.Context, repo *RecipeRepository) *RecipeDe
...
-		return nil
-	}
+func (dt *RecipeDetector) GetDetectedRecipes() (map[string]*RecipeDetectionResult,
+	map[string]*RecipeDetectionResult,
+	error) {
 
-	dt.isDetectionRunned = true
+	if len(dt.availableRecipes) != 0 || len(dt.unavaliableRecipes) != 0 {
+		return dt.availableRecipes, dt.unavaliableRecipes, nil
+	}
```
mbruzina, line 51: How could these two maps already have recipes?  `GetDetectedRecipes` only seems to be called from one place and appears responsible for setting these maps 🤔

NRhzhao: Updated

mbruzina (dismissed):

internal/install/recipe_installer.go
```diff
@@ -288,14 +290,45 @@ func (i *RecipeInstall) install(ctx context.Context) error {
...
+		}
+	}
+	fmt.Println(message)
+}
+
+func (i *RecipeInstall) reportRecipeStatuses(availableRecipes map[string]*recipes.RecipeDetectionResult,
+	unavailableRecipes map[string]*recipes.RecipeDetectionResult) {
+
+	for _, d := range unavailableRecipes {
+		e := execution.RecipeStatusEvent{Recipe: *d.Recipe, ValidationDurationMs: d.DurationMs}
```
Julien4218, line 308: @NRhzhao I'm not sure we'd want to do that. This will insert an event (which status?) for every recipe that doesn't detect? Or is this reporting the recipes that are detected but not available? If those are the "detected/not available recipes" perhaps we could rename unavailableRecipes to detectedRecipes, or detectedUnavailableRecipes?

internal/install/recipe_installer.go
```diff
@@ -288,14 +290,45 @@ func (i *RecipeInstall) install(ctx context.Context) error {
...
+		i.status.ReportStatus(d.Status, e)
+	}
+
+	// Sort the key, so we can get consistent ordering
+	keys := make([]string, 0)
+	for k := range availableRecipes {
+		keys = append(keys, k)
+	}
+	sort.Strings(keys)
```
Julien4218, line 317: Is this sort following the desired order? Infra first, then log...?

Julien4218 (commented):
I left a couple of comments

internal/install/recipe_installer.go
NRhzhao: It doesn't report the status of the recipes that are not detected.  I believe this is the same way we did detection for the guided install before.  I double-checked the log, and it doesn't produce any extra events.
This doesn't just report unavailable recipes, it also reports available recipes.  The use case is if a person does a target install for PHP, and we detected MySQL on the same host and status is available.  MySQL would be part of the available recipes.  So we want to also report the available recipes in case, so we just report detect status here for datagap.  During installation, the other events for the recipe that's installed will be handled correctly.

internal/install/recipe_installer.go
NRhzhao: We use the map here since we use available recipes for lookup in the bundler.  We can have the detector return an ordered slice, and I can update the bundler to have it convert to a map.  
Let me know if you think that's a better appraoch.

NRhzhao: Updated to have stable sort.

internal/install/recipe_installer.go
Julien4218: TY

PR #1306, opened 2022-05-23
chore(install): Allow user continue to install even when core bundle

NRhzhao:
APM can still be installed if the core bundle failed to install.  We want to allow users to continue attempt installing APM if the core bundle failed.


NRhzhao:
Close, for now, the risk outweighs the benefit will review it later.

PR #1309, opened 2022-05-31
fix(install): pre-install eval panic should return valid result

NRhzhao:


NRhzhao:
> @NRhzhao this looks good approving, however I think the ShouldPanic is not used and should be removed.

Hi Julien, I added the flag for the mock so we can have a test that panic when run.

PR #1365, opened 2022-09-23
README: Update "developer toolkit" link to developers.newrelic.com

lewisdoesstuff:
The link previously referenced a github.io link that's currently a 404 (and has been for a while). I couldn't find a mirror of this, so linking to https://developer.newrelic.com seems like the best alternative.


kidk:
Thanks @lewisdoesstuff ❤️

PR #1366, opened 2022-10-04
chore(install): create simple test script + dockerfile example

kidk:
Just some testing and open install fun, ignore for now.


open-install-test.sh
```diff
@@ -0,0 +1,64 @@
...
+CHANGES=$(git status --porcelain | wc -l)
+if [ "$CHANGES" -gt 0 ]; then
+    make compile-linux &
+    pids[0]=$!
+    make compile-darwin &
+    pids[1]=$!
+fi
+
+# Build container async
+docker build -t samuel:test . &
```
Julien4218, line 42: @kidk maybe rename samuel:test to local:test or something else?

open-install-test.sh
```diff
@@ -0,0 +1,64 @@
...
+  echo "Please set the NEW_RELIC_REGION environment variable"
+  exit 1
+fi
+
+# Check if we have the open install repo locally
+if [ ! -d "tmp/" ]
+then
+    mkdir tmp/ || true
+    pushd tmp
+    git clone git@github.com:newrelic/open-install-library.git
```
Julien4218, line 24: alternatively, you can just use "make recipes" to fetch the latest recipe release.

PR #1382, opened 2022-10-28
feat: add entity changetracker create command

mbazhlekova:


NSSPKrishna (dismissed):

docs/cli/newrelic_entity_changetracker_create.md
```diff
@@ -0,0 +1,52 @@
...
+```
+newrelic entity changetracker create --guid <GUID> --version <1.0.0>
+```
+
+### Options
+
+```
+  -h, --help                    help for create
+  -g, --guid string             the entity GUID to create change tracker
+  -v, --version string          the tag names to add to the entity
```
jonathanpdx, line 27: For `version` and the other fields, perhaps we should copy as close as practical [the docs that we have for the graphql api](https://source.datanerd.us/markers/deployment-marker-api/blob/2da074e90d835b3d01f4be021294e95d4aa57fe9/src/main/resources/schema.graphql#L29-L50)?

> The version of the deployed software, for example, something like v1.1

jonathanpdx: Also, can we indicate that `version` is required? And include the necessary validation that it's a non-blank value?

PR #1384, opened 2022-11-01
feat(logs): removed usage of temp file for log output

mbruzina:
Removed capturing CLI output in temp file in favor of storing in memory.  This removes unnecessary file handling/cleanup that surfaced some issues with our windows infra recipes last week

### DIFF
--- a/internal/install/execution/output_parser.go
+++ b/internal/install/execution/output_parser.go
@@ -33,9 +33,13 @@ func (op *OutputParser) Metadata() map[string]string {
 	return nil
 }
 
-func (op *OutputParser) FailedRecipeOutput() string {
-	if val, ok := op.output["FailedRecipeOutput"]; ok {
-		return val.(string)
+func (op *OutputParser) IsCapturedCliOutput() bool {
+	if val, ok := op.output["CapturedCliOutput"]; ok {
+		valAsBool, isBool := val.(bool)
+		if !isBool {
+			return false
+		}
+		return valAsBool
 	}
-	return ""
+	return false
 }
--- a/internal/install/execution/recipe_log_forwarder.go
+++ b/internal/install/execution/recipe_log_forwarder.go
@@ -18,6 +18,11 @@ import (
 	"github.com/newrelic/newrelic-client-go/v2/pkg/region"
 )
 
+type LogForwarder interface {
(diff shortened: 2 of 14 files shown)


internal/install/execution/output_parser.go
```diff
@@ -33,9 +33,13 @@ func (op *OutputParser) Metadata() map[string]string {
 	return nil
 }
 
-func (op *OutputParser) FailedRecipeOutput() string {
-	if val, ok := op.output["FailedRecipeOutput"]; ok {
-		return val.(string)
+func (op *OutputParser) LogCaptureEnabled() bool {
```
Julien4218, line 36: @mbruzina small feedback about naming and consistency. It would be great to use a similar/identical naming between the var and its usage (reduces reader confusion when browsing across repositories). 
Also for boolean returned method, it usually reads better if the spelling is turned into a question, for example "IsLogCaptureEnabled". Otherwise this could be interpreted by a reader as "make the log capture enabled", or "log the enabled capture"

internal/install/execution/recipe_log_forwarder.go
```diff
@@ -51,15 +56,8 @@ func (lf *RecipeLogForwarder) PromptUserToSendLogs(reader io.Reader) bool {
...
-func (lf *RecipeLogForwarder) SendLogsToNewRelic(outputFilePath string, recipeName string) {
-	// open file, build log entries
-	f, err := os.Open(outputFilePath)
-	if err != nil {
-		log.Fatalf("open file error: %v", err)
-		return
-	}
-	defer f.Close()
-	lf.buildLogEntryBatch(recipeName, bufio.NewScanner(f))
+func (lf *RecipeLogForwarder) SendLogsToNewRelic(recipeName string, recipeOutput []string) {
```
Julien4218, line 59: @mbruzina I like this method name, very clear. I would just rename the recipeOutput parameter as it seems to tie a specific use case to what that function can do, when it doesn't matter what kind of string array is passed in (it doesn't have to be a recipe output string array).

Julien4218 (approved):
Good work.

mbruzina:
Updates completed per feedback cc @amolero-nr @Julien4218

Julien4218 (approved):

PR #1386, opened 2022-11-02
feat(log-metadata): adding SendLogsOptIn recipe metadata 

mbruzina:
Adding functionality to flag recipes in which users have opted in to send logs to New Relic via our LogsApi.

If a recipe installation fails, has logging output in-memory and contains the metadata `CapturedCliOutput` = `true`,  we will prompt the user requesting permission to send logs.

Once the user replies y/n, we add a new key to the recipe metadata `SendLogsOptIn` with a true/false value prior to posting the recipe event to install-events-service.  

this cli update should be released in tandem with the following open-install-library PR: https://github.com/newrelic/open-install-library/pull/811


NRhzhao:
1. Tested assumeYes = true, and didn't get the prompt when infra failed (Expected?)
2. Tested assumeYes = false, did get the prompt when infra failed
3. Tested infra installed, no prompt

Not sure where the log is being stored, but everything else looks good to me

NRhzhao (dismissed):

NRhzhao (dismissed):

mbruzina:
@NRhzhao assumeYes implies the silent install, in which we aren't capturing stdout/stderr - this should never prompt.  `assumeYes=false` or no assumeYes/y flag makes the install verbose and we would capture...only prompting if a given recipe is flagged to capture logs _and_ the installation failed

amolero-nr (approved):

PR #1389, opened 2022-11-03
feat(install): Allow tag to be passed in via install command

NRhzhao:


NRhzhao:
This allows `--tag` flag to be used with the `install` command.
1.  This will support backward compatibility with NRIA_CUSTOM_ATTRIBUTES. It will merge and generate YAML if both configs are defined.
2. Following are tested and working commands
With both defined:
`NRIA_CUSTOM_ATTRIBUTES='{"install_from":"virtuoso"}' ./newrelic install --tag tag3:2345,tag4:abb,made:hzhao -y`
With just tags:
`./newrelic install --tag justTags:yes`

Help Info in CLI:
image

The expected result is custom attribute section is added to infra config:
`cat /etc/newrelic-infra.yml`
image

And you can see the tags in NR1 as well:
image

mbruzina (approved):

internal/install/command.go
```diff
@@ -21,6 +21,7 @@ var (
 	recipeNames  []string
 	recipePaths  []string
 	testMode     bool
+	entityTags   []string
```
Julien4218, line 24: @NRhzhao I would have simple name this tags, as those could apply to various things, maybe not entity in the sense of NR1 entity.

Julien4218 (commented):
@NRhzhao this is fine work. I would only suggest renaming all the entity tag names to just tag.

internal/install/command.go
NRhzhao: Thanks @Julien4218  Will make the update

amolero-nr (approved):

PR #1417, opened 2023-01-25
feat(install): wire deployedBy attribute when installing

Julien4218:


Julien4218:
This is blocked until we can get an updated newrelic-client-go with the latest tutone generation https://github.com/newrelic/newrelic-client-go/actions/workflows/generate.yml

internal/install/types/install_context.go
```diff
@@ -18,7 +21,8 @@ type InstallerContext struct {
 	RecipePaths []string
 	// LocalRecipes is the path to a local recipe directory from which to load recipes.
 	LocalRecipes string
-	Tags         []string
+	// Tags         []string
```
rthorn-nr, line 24: Is this meant to be commented out? I would think this either should be included or deleted

Julien4218: Good catch, I forgot to remove as I was working on the implementation and test.

amolero-nr (approved):

PR #1421, opened 2023-02-01
fix(scripts): install dir check

amolero-nr:
If it does not exist, creates default `DESTDIR` path (`/usr/local/bin`) where the `install.sh` script moves the downloaded `newrelic` binary to. Corrects `not a directory` error in MacOS.


sanderblue (approved):

kidk (approved):
LGTM

Julien4218 (approved):

PR #1427, opened 2023-02-14
Fixed usage description for `yq` and added documentation.

pnvnd:


internal/utils/command_yq.go
```diff
@@ -26,8 +26,8 @@ var (
 
 var cmdYq = &cobra.Command{
 	Use:   "yq",
-	Short: "Parse json strings",
-	Long: `Parse json strings
+	Short: "Parse yaml strings",
+	Long: `Parse yaml strings
 
 The yq subcommand makes use of gojq (https://github.com/itchyny/gojq) to provide
```
kidk, line 32: Not sure about this. @sanderblue Do you know if that's correct?

kidk:
Thanks a lot for the PR @pnvnd 

I've asked to double check on thing, and we need to look at failing tests, but your changes look good. Thanks a lot

internal/utils/command_yq.go
mbruzina: @kidk This is correct, it seems we copied contents from `command_jq` when yq was added, just missed this

mbruzina:
@kidk This failure may be because we're merging from a forked repository and the fork doesn't have all of the required secrets for our actions to run.  The failure is due to a missing `PGP_PRIVATE_KEY` envar; these are present in our non-forked PR action runs.  Considering these are all content changes, we can ignore/merge IMO if that's alright with you

PR #1432, opened 2023-02-23
Improve proxy support for install scripts

rthorn-nr:
For `install.sh` and `install.ps1`, check for connectivity to `https://download.newrelic.com` and provide a helpful message if unable to connect.

For `install.ps1`, adds a proxy to the requesting `WebClient` via the optional environment variable `HTTPS_PROXY`, which will enable setting a proxy for the CLI download.

### DIFF
--- a/scripts/install.ps1
+++ b/scripts/install.ps1
@@ -3,6 +3,17 @@ if (!$p.IsInRole([System.Security.Principal.WindowsBuiltInRole]::Administrator))
     throw 'This script requires admin privileges to run and the current Windows PowerShell session is not running as Administrator. Start Windows PowerShell by using the Run as Administrator option, and then try running the script again.'
 }
 [Net.ServicePointManager]::SecurityProtocol = 'tls12, tls';
-$version = (New-Object System.Net.WebClient).DownloadString("https://download.newrelic.com/install/newrelic-cli/currentVersion.txt").Trim();
-(New-Object System.Net.WebClient).DownloadFile("https://download.newrelic.com/install/newrelic-cli/${version}/NewRelicCLIInstaller.msi", "$env:TEMP\NewRelicCLIInstaller.msi");
+$WebClient = New-Object System.Net.WebClient
+if ($env:HTTPS_PROXY) {
+    $WebClient.Proxy = New-Object System.Net.WebProxy($env:HTTPS_PROXY, $true)
+}
+$version = $null
+try {
+    $version = $WebClient.DownloadString("https://download.newrelic.com/install/newrelic-cli/currentVersion.txt").Trim();
+    $WebClient.DownloadFile("https://download.newrelic.com/install/newrelic-cli/${version}/NewRelicCLIInstaller.msi", "$env:TEMP\NewRelicCLIInstaller.msi");
+}
+catch {
+    Write-Output "`nCould not download the New Relic CLI installer.`n`nCheck your firewall settings. If you are using a proxy, make sure that you are able to access https://download.newrelic.com and that you have set the HTTPS_PROXY environment variable with your full proxy URL.`n"
+    throw
+}
 msiexec.exe /qn /i $env:TEMP\NewRelicCLIInstaller.msi | Out-Null;
--- a/scripts/install.sh
+++ b/scripts/install.sh
@@ -60,8 +60,12 @@ if [ $IS_CURL_INSTALLED -eq 0 ]; then
 fi
 
 # GitHub's URL for the latest release, will redirect.
-LATEST_URL="https://download.newrelic.com/install/newrelic-cli/currentVersion.txt"
+BASE_URL="https://download.newrelic.com"
+LATEST_URL="$BASE_URL/install/newrelic-cli/currentVersion.txt"
 DESTDIR="${DESTDIR:-/usr/local/bin}"
+
+# Check for connectivity to https://download.newrelic.com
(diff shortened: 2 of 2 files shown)


scripts/install.ps1
```diff
@@ -1,8 +1,19 @@
...
-(New-Object System.Net.WebClient).DownloadFile("https://download.newrelic.com/install/newrelic-cli/${version}/NewRelicCLIInstaller.msi", "$env:TEMP\NewRelicCLIInstaller.msi");
+$WebClient = New-Object System.Net.WebClient
+if ($proxyUrl) {
+    $WebClient.Proxy = New-Object System.Net.WebProxy($proxyUrl, $true)
+}
+$version = $null
+try {
+    $version = $WebClient.DownloadString("https://download.newrelic.com/install/newrelic-cli/currentVersion.txt").Trim();
+}
+catch {
```
Julien4218, line 15: It might be good to get the exception, and add the inner message below.

Julien4218: We should test this as I recall some weirdness sometimes with powershell and try/catch not catching all exceptions.

Julien4218: Is there a way to put a connect timeout? I don't think it should be more than 10s, just like the linux shell version.

scripts/install.ps1
```diff
@@ -1,8 +1,19 @@
+param ($proxyUrl)
 $p = New-Object System.Security.Principal.WindowsPrincipal([System.Security.Principal.WindowsIdentity]::GetCurrent())
 if (!$p.IsInRole([System.Security.Principal.WindowsBuiltInRole]::Administrator)) {
     throw 'This script requires admin privileges to run and the current Windows PowerShell session is not running as Administrator. Start Windows PowerShell by using the Run as Administrator option, and then try running the script again.'
 }
 [Net.ServicePointManager]::SecurityProtocol = 'tls12, tls';
-$version = (New-Object System.Net.WebClient).DownloadString("https://download.newrelic.com/install/newrelic-cli/currentVersion.txt").Trim();
-(New-Object System.Net.WebClient).DownloadFile("https://download.newrelic.com/install/newrelic-cli/${version}/NewRelicCLIInstaller.msi", "$env:TEMP\NewRelicCLIInstaller.msi");
+$WebClient = New-Object System.Net.WebClient
+if ($proxyUrl) {
```
Julien4218, line 8: Is there a reason why we wouldn't use the same HTTPS_PROXY mechanism for windows?

Julien4218 (commented):
@rthorn-nr I left a few comments.

scripts/install.ps1
rthorn-nr: No specific reason other than thinking that it would be more flexible to pass it as a script parameter than to set the user environment, but also no reason to break from convention. Pushing up a change for this momentarily

scripts/install.ps1
rthorn-nr: > Is there a way to put a connect timeout?

At first glance, it doesn't look like there's a way to do this with [WebClient](https://learn.microsoft.com/en-us/dotnet/api/system.net.webclient?view=net-7.0) (unlike the newer [HttpClient](https://learn.microsoft.com/en-us/dotnet/api/system.net.http.httpclient?view=net-7.0), which has a timeout property). There's a [workaround](https://stackoverflow.com/a/12879118) which involves extending the class, but I'm seeing a timeout after about 20s... if we want to get that down to 10s, it'll just take some reworking one way or another, so I'd leave it up to you whether we want to shorten the feedback loop.

rthorn-nr: Going to do some testing with exceptions and get back to you on those comments

Julien4218: I was going to suggest to keep what we have in regard of the timeout, but comparing the requirements for each of the method (apply to at the bottom), I noticed the HttpClient has broader support than WebClient. Also the doc on WebClient mentions "We don't recommend that you use the WebClient class for new development. Instead, use the [System.Net.Http.HttpClient](https://learn.microsoft.com/en-us/dotnet/api/system.net.http.httpclient?view=net-7.0) class."
I recall using the HttpClient in the past, I believe it's a more bare bone way to make request. It might be good to switch to using it actually, and avoid compatibility issues for hosts that don't have the right .NET version.

rthorn-nr: Error output should be improved by the latest [commit](https://github.com/newrelic/newrelic-cli/pull/1432/commits/766fef757378d0f11471f76f9382ab3d2a42ce26). Moved the file download to the try/catch as well. 

I'd naively assume that a regular try/catch would catch all exceptions coming out of those two download commands -- let me know how I can help test that powershell try/catch weirdness

rthorn-nr: Missed your comment above yesterday. Makes sense, happy to refactor to use HttpClient

Julien4218 (dismissed):
Approving after verifying on linux and windows with proxy.

scripts/install.ps1
rthorn-nr: @Julien4218 I think we should stick with WebClient for now. 

Testing using PS v5.1 (default for Server 2019 and Server 2022), HttpClient's async behavior has proven hard to handle in a PS script. Using a strategy like [this](https://stackoverflow.com/a/51218694) and calling `GetResult()` off of `GetAwaiter()` cancels the request, and the `await` syntax is not available. 

I also tried using `Invoke-WebRequest`, but that triggered an IE security warning: `Content within this application coming from the website listed below is being blocked by Internet Explorer Enhanced Security Configuration: about:security_powershell.exe`. Might just by my testing setup, but using WebClient does not trigger this warning.

It's a shame, especially since HttpClient automatically reads proxy variables from the environment, as does `Invoke-Webrequest` in PS versions 7+. Seems to me like it's still a bit of a compatibility risk to update this script in that way.

Julien4218 (approved):

PR #1433, opened 2023-02-28
fix(install): Update scripts/install.sh to use a proxy with curl if HTTPS_PROXY is set

steve-cunnew:
The updates I have made will detect if the HTTPS_PROXY environment variable has been set, and if it has it will then use the proxy with the subsequent curl commands (check connection, downloaded latest version info, then download the correct tar file). This is required so that the guided install can then be updated to set the proxy environment variable in the command line that users will copy and paste from the UI, allowing it to work when a proxy must be used from a host.

### DIFF
--- a/scripts/install.sh
+++ b/scripts/install.sh
@@ -59,11 +59,9 @@ if [ $IS_CURL_INSTALLED -eq 0 ]; then
     fi
 fi
 
-# Set curl flags to use the proxy if one has been specified 
+# Set curl proxy flag to use HTTPS_PROXY if it has been set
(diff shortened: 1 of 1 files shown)


NRhzhao:
@rthorn-nr Can you help review and test this.

rthorn-nr:
Hi @steve-cunnew , `curl` will apply the `HTTPS_PROXY` variable if it is set in the shell environment, so there's no need to explicitly set it for these commands. 

I'm going to close this PR, but please open a new PR if there are further changes you'd like to see to improve our proxy support.

PR #1438, opened 2023-03-22
chore(install): Improve network error detection and messaging

rthorn-nr:
Moves the call to `checkNetwork()` from inside the installer to before fetching the license key. Also adds more descriptive and helpful error messages if `checkNetwork()` fails.

### DIFF
--- a/internal/install/recipe_installer.go
+++ b/internal/install/recipe_installer.go
@@ -55,8 +55,6 @@ type RecipeInstall struct {
 type RecipeInstallFunc func(ctx context.Context, i *RecipeInstall, m *types.DiscoveryManifest, r *types.OpenInstallationRecipe, recipes []types.OpenInstallationRecipe) error
 
 func NewRecipeInstaller(ic types.InstallerContext, nrClient *newrelic.NewRelic) *RecipeInstall {
-	checkNetwork(nrClient)
-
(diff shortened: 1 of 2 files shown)


mbruzina (dismissed):

internal/install/recipe_installer.go
```diff
@@ -718,17 +716,24 @@ func (i *RecipeInstall) finishHandlingFailure(recipeName string) {
...
+func checkNetwork(nrClient *newrelic.NewRelic) error {
 	err := nrClient.TestEndpoints()
 	if err != nil {
 
 		proxyConfig := httpproxy.FromEnvironment()
 
 		log.Debugf("proxyConfig: %+v", proxyConfig)
 		if proxyConfig.HTTPProxy != "" || proxyConfig.HTTPSProxy != "" || proxyConfig.NoProxy != "" {
-			log.Warn("Proxy settings have been configured but we are still unable to connect to the New Relic platform.  You may need to adjust your proxy environment variables.  https://github.com/newrelic/newrelic-cli/blob/main/docs/GETTING_STARTED.md#using-an-http-proxy")
+			log.Warn("Proxy settings have been configured, but we are still unable to connect to the New Relic platform.")
```
NRhzhao, line 727: It might be helpful to output the current proxy config here so it's easy to check if they have the correct settting?

PR #1447, opened 2023-04-14
chore(install): Remove user/account NerdStorage writes

stevula:
## Summary

Removes writes to account- and user-scoped NerdStorage from the install status reporter. The status reporter should only write to entity scope.

## Context

New Relic UI no longer depends on the `openInstallLibrary` collection, except for entity-scoped NerdStorage data.

### DIFF
--- a/internal/install/execution/nerdstorage_status_reporter_integration_test.go
+++ b/internal/install/execution/nerdstorage_status_reporter_integration_test.go
@@ -46,10 +46,8 @@ func TestReportRecipeSucceeded_SingleEntityGUID(t *testing.T) {
 	status := NewInstallStatus(types.InstallerContext{}, []StatusSubscriber{r}, NewPlatformLinkGenerator())
 	status.withEntityGUID(entityGUID)
 
-	defer deleteUserStatusCollection(t, c.NerdStorage)
 	defer deleteEntityStatusCollection(t, entityGUID, c.NerdStorage)
 	defer deleteEntity(t, entityGUID, c)
-	defer deleteAccountStatusCollection(t, a, c.NerdStorage)
 
 	rec := types.OpenInstallationRecipe{Name: "testName"}
 	evt := RecipeStatusEvent{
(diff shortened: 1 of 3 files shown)


stevula:
It looks like the codecov check is failing because the coverage for [mock_nerdstorage_client.go](https://app.codecov.io/gh/newrelic/newrelic-cli/pull/1447/blob/internal/install/execution/mock_nerdstorage_client.go) dropped by 37.50%. This an optional check I think? 

Code coverage went down because the mock client's methods `WriteDocumentWithUserScope` and `WriteDocumentWithAccountScope` are no longer referenced in any tests. There's a possibility they might be useful in testing other parts of the code that deal with user/account storage, so I didn't remove them. Doesn't seem too worrisome to have some testing coverage gaps on a mock client 🤷

stevula:
integration tests failing. looking into it 👀

stevula:
I pushed a minor change and then reverted it. Now integration tests are passing again (I was pretty sure they were passing originally before I rebased on main). I think these tests might be a bit flakey for some reason.

internal/install/execution/nerdstorage_status_reporter_integration_test.go
```diff
@@ -135,26 +140,27 @@ func getEntityStatusCollection(t *testing.T, guid string, c nerdstorage.NerdStor
...
+// TODO: remove after 5/20/23 since we are no longer writing to these collections.
 func deleteUserStatusCollection(t *testing.T, c nerdstorage.NerdStorage) {
 	di := nerdstorage.DeleteCollectionInput{
 		Collection: collectionID,
 		PackageID:  packageID,
 	}
-	ok, err := c.DeleteCollectionWithUserScope(di)
+	_, err := c.DeleteCollectionWithUserScope(di)
 	require.NoError(t, err)
-	require.True(t, ok)
```
stevula, line 155: ok can be false if the collection doesn't exist but it doesn't matter. This is just to clean up nerdstorage from previous test runs.

NRhzhao (approved):

sanderblue (approved):

PR #1448, opened 2023-04-14
chore(install): Suppress output JSON warning for empty strings

rthorn-nr:


internal/install/execution/go_task_recipe_executor.go
NRhzhao, line 1: maybe just do i

f line != "" && json.Valid([]byte(line)) {

rthorn-nr: That was my first thought too, but that sends us to that else block with the debug message

NRhzhao:
Another PR opened for this.

PR #1449, opened 2023-04-14
chore(install): Refactor re.setOutput to use bufio scanner

rthorn-nr:
Also fixes invalid JSON warning message on empty string caused by newline in output. Alternative to https://github.com/newrelic/newrelic-cli/pull/1448


internal/install/execution/go_task_recipe_executor.go
```diff
@@ -207,7 +207,7 @@ func (re *GoTaskRecipeExecutor) setOutput(outputFileName string) {
 		}
 	}
 	if err = s.Err(); err != nil {
-		log.Debugf("error scanning output file %s: %w", outputFileName, err)
```
amolero-nr: `%w` is supported: https://go.dev/blog/go1.13-errors

amolero-nr (dismissed):

internal/install/execution/go_task_recipe_executor_test.go
```diff
@@ -486,3 +484,67 @@ tasks:
 	assert.Empty(t, e.GetOutput().Metadata()["something"])
 	assert.Equal(t, "very", e.GetOutput().Metadata()["clean"])
 }
+
+func Test_mergeStringJSON(t *testing.T) {
```
Julien4218, line 488: This is a long test, perhaps we could break it into many small tests?

Julien4218 (commented):
Maybe rework the test to be more focus and smaller.

PR #1454, opened 2023-05-01
chore(install): increase license key fetch retry timeout to 5 minutes

sanderblue:


internal/install/command.go
```diff
@@ -98,7 +98,8 @@ func assertProfileIsValid() error {
 		return fmt.Errorf("region is required")
 	}
 
-	licenseKey, err := client.FetchLicenseKey(accountID, config.FlagProfileName)
+	maxRetries := 60
+	licenseKey, err := client.FetchLicenseKey(accountID, config.FlagProfileName, &maxRetries)
```
stevula, line 102: Something to consider: this assumes a global default retry delay of 5s. If the default retry delay changes, the client will not be retrying for 5 minutes. If we want to guarantee 5 minutes, we could either:
1. Set max retries and delay explicitly (`maxRetries = 60, delay = 5000`)
2. Calculate max retries backwards from 5 minutes `maxRetries = 300_000 / delay`

sanderblue: Fair point, I'll see what adjustments I can make.

internal/config/config.go
```diff
@@ -27,8 +27,9 @@ const (
 	CredentialsFileName    = "credentials.json"
 	DefaultPluginDir       = "plugins"
 
-	DefaultPostRetryDelaySec = 5
-	DefaultPostMaxRetries    = 20
+	DefaultPostRetryDelaySec  = 5
```
Julien4218, line 30: @sanderblue where else are we using those values to retry less than 5min? If it's anywhere related to installation, we should probably update those values as well to retry up to 5 min.

sanderblue: One other place we use retries is during the [`ConfigValidator.Validate()` method](https://github.com/newrelic/newrelic-cli/blob/main/internal/diagnose/config_validator.go#L72-L86) for the `postEvent` when calling `client.Events.CreateEventWithContext()`.

Do we also want to increase that timeout? Or are we only concerned with the license key fetch timeout?

Julien4218: I think we should have a consistent retry/loop and consolidate on 5min max timeout. Otherwise, I'd be concerned we would have other flows that will experience a faster timeout when the account is too new.

sanderblue: One thing to note, if we do increase the timeouts for all of these retry methods, it could increase the total installation time fairly dramatically if the timeouts are hit since we call these methods multiple times throughout the installation process.

PR #1458, opened 2023-05-04
feat(install): add event tracking with segment

NRhzhao:
https://issues.newrelic.com/browse/NR-114517


internal/install/command.go
```diff
@@ -41,17 +42,26 @@ var Command = &cobra.Command{
 		logLevel := configAPI.GetLogLevel()
 		config.InitFileLogger(logLevel)
 
-		if err := checkNetwork(client.NRClient); err != nil {
+		sg := segment.New()
+
+		err := assertProfileIsValid(config.DefaultMaxTimeoutSeconds, sg)
```
Julien4218, line 47: @NRhzhao it looks like we may get an error below, could we start with an event to indicate we're starting the cli init?
Also could we get the accountID and region from the env var? That would be valuable information in case we get any of the errors afterward.

internal/install/command.go
```diff
@@ -84,39 +94,43 @@ var Command = &cobra.Command{
...
+		return err
+	}
+
 	licenseKey, err := client.FetchLicenseKey(accountID, config.FlagProfileName, &maxTimeoutSeconds)
 	if err != nil {
-		return fmt.Errorf("could not fetch license key for account %d: %s", accountID, err)
+		sg.Track(accountID, segment.NewEvent(segment.EventTypes.UnableToFetchLicenseKey, err.Error()))
+		return fmt.Errorf("could not fetch license key for account %d:, license key: %v %s", accountID, utils.Obfuscate(licenseKey), err)
 	}
+	sg.Track(accountID, segment.NewEvent(segment.EventTypes.AbleToFetchLicenseKey, ""))
```
Julien4218, line 127: @NRhzhao it might be good to use a different syntax for that event type so it doesn't cover the syntax for the Unable event. Maybe "LicenseKeyFetchedOk"

internal/install/segment/client.go
```diff
@@ -0,0 +1,121 @@
...
+
+func NewEvent(event EventType, detail string) Event {
+	return Event{
+		event,
+		detail,
+	}
+}
+
+func getWriteKey() (string, error) {
+	data, err := embedded.ReadFile("files/events.src")
```
Julien4218, line 114: @NRhzhao I think this should be done somewhere else and injected on init to avoid the coupling with Embedded resource at this level.

internal/install/segment/client.go
```diff
@@ -0,0 +1,121 @@
...
+	}
+
+	properties := toMap(event)
+
+	properties["category"] = "newrelic-cli"
+	properties["accountId"] = accountID
+
+	err := client.Enqueue(analytics.Track{
+		UserId:     fmt.Sprintf("%d", accountID),
+		Event:      eventName,
```
Julien4218, line 76: @NRhzhao I think there is a mistake here, it should be the eventName passed in from the method parameter. Also, I don't think that const eventName is needed.

internal/install/segment/client.go
```diff
@@ -0,0 +1,121 @@
...
+var (
+	embedded embed.FS
+)
+
+type Segment struct {
+	analytics.Client
+}
+
+func New() *Segment {
+	writeKey, err := getWriteKey()
```
Julien4218, line 47: We should handle the case when we have no write key defined, and I don't think this should even be consider a warning. I would just make that a debug statement.

internal/install/segment/client.go
```diff
@@ -0,0 +1,121 @@
...
+		log.Warnf("segment: error reading write key, cannot write to segment %v", err)
+		return nil
+	}
+	if writeKey == "" {
+		log.Warn("segment: write key is empty, cannot write to segment")
+		return nil
+	}
+
+	client := analytics.New(writeKey)
+	log.Info("segment initialized")
```
Julien4218, line 58: @NRhzhao please let's not output this to the user. Maybe debug, or no log at all since this is the main path. As long as we log when we're in the unexpected cases, I don't think we need to log that one.

NRhzhao: Will remove

internal/install/command.go
NRhzhao: All the tracking is done inside the assertProfileIsValid function since we want to have specific event type fo different errors that occurred.

internal/install/segment/client.go
```diff
@@ -0,0 +1,95 @@
...
+	}
+
+	properties := toMap(event)
+
+	properties["category"] = "newrelic-cli"
+	properties["accountId"] = accountID
+
+	err := client.Enqueue(analytics.Track{
+		UserId:     fmt.Sprintf("%d", accountID),
+		Event:      "VirtuosoCLIInstall",
```
Julien4218, line 60: @NRhzhao check IES, this isn't right. the Event field should be the eventName.
The method signature should allow an optional interface{ } to get additional properties.

internal/install/segment/client.go
```diff
@@ -0,0 +1,95 @@
...
+	UnableToFetchLicenseKey: "UnableToFetchLicenseKey",
+	LicenseKeyFetchedOk:     "LicenseKeyFetchedOk",
+	UnableToOverrideClient:  "UnableToOverrideClient",
+}
+
+type Segment struct {
+	analytics.Client
+}
+
+func New(writeKey string) *Segment {
```
Julien4218, line 37: @NRhzhao what if we're injecting the additional fields we always want to track event during init, for example accountID and region? Could we maybe get them from os.GetEnv( ). Note for region, we should also handle the case when empty means US.
That will make analyzing those events quite easier.
This will also lighten up the contract of the Track( ) method.

internal/install/segment/client.go
NRhzhao: event name is the table name, in IES there are two event names that correspond to the two tables we have.
I figure for CLI, we want only one table so I hardcode the one table name there.
In IES, the event is a interface type, so it can be any type and gets converted to a dictionary. 
Since we only logging event here, I thought to keep it as a concrete type instead.

image

Julien4218 (approved):

PR #1463, opened 2023-05-09
fix(install): segment flushing issue

NRhzhao:
Segment uses an internal queue, when the CLI exit early, the queued message does not get sent.

Segment doc suggests setting the batching size of 1 to immediate send, but that didn't seem to work by it self.

It also suggests using close, which calls for flushing on the queue, but that also didn't seem to work.

In the end, I have to add a manual 5-second wait for the flush to complete, which seems to be working now.


internal/install/command.go
```diff
@@ -55,7 +54,12 @@ var Command = &cobra.Command{
 		c, err := client.NewClient(configAPI.GetActiveProfileName())
 		if err != nil {
 			// An error was encountered initializing the client.  This may not be a
-			// problem since many commands don't require the use of an initialized client
+			// problem since many commands don't require the use of an initialized client			sg := initSegment()
+			sg := initSegment()
+			defer func() {
+			  sg.Close()
+				time.Sleep(5 * time.Second)
```
Julien4218, line 61: @NRhzhao this seems excessive, I don't think a user is going to wait 5s if they already cancelled the execution. Isn't there a way to trigger a synchronous flush? Wouldn't the close do that? Otherwise I would not recommend waiting more than 1s at most.

NRhzhao: @Julien4218 I run the test, and it does wait even in context cancel. The segment runs on a 5-second timer to flush, and I can reduce the timer, but run the risk of losing the message.

Julien4218 (approved):
Approved for now given the limitation of the library (no way to ensure the flush is done). We should look into invoking the http api directly in a follow up ticket.

PR #1465, opened 2023-05-12
chore: add more segment logs

NRhzhao:


NRhzhao:
1. Move the error type from diagnostics to types
2. Try to do all segment logging inside the segment reporter, but the recipe install requires NR client in its initialization. The assert profile checks for licensing updates the NR client to use the correct API key, so cannot move the assert licensing into recipe installer

internal/install/types/errors.go
```diff
@@ -113,3 +186,41 @@ func (e *UncaughtError) Error() string {
...
 var ErrorFetchingLicenseKey = errors.New("Oops, we're having some difficulties fetching your license key. Please try again later, or see our documentation for installing manually https://docs.newrelic.com/docs/using-new-relic/cross-product-functions/install-configure/install-new-relic")
+
+type ErrUnalbeToFetchLicenseKey struct {
+	Err     error
+	Details string
+}
+
+func (e *ErrUnalbeToFetchLicenseKey) Error() string {
+	return "could not fetch license key"
+}
```
rthorn-nr, line 197: Typo: `ErrUnalbeToFetchLicenseKey` > `ErrUnableToFetchLicenseKey`

Which brings up another question: is this type referenced anywhere?

NRhzhao: This is actually existing behavior, I just moved the code

rthorn-nr (approved):

PR #1467, opened 2023-05-20
fix(install): swap segment client to use http api

Julien4218:


NRhzhao (approved):
I tested with Alex, and it's still spotty receiving the messages, but it doesn't seem worst than the current implementation. We think it could be a UI bug on the segment end.

Approved

PR #1469, opened 2023-05-23
chore(install): Skip OTEL if not targeted

rthorn-nr:
Preparing for new OTEL recipe for targeted install only

### DIFF
--- a/internal/install/recipe_installer_test.go
+++ b/internal/install/recipe_installer_test.go
@@ -168,6 +168,33 @@ func TestInstallGuidedShouldNotSkipCoreInstall(t *testing.T) {
 	assert.Equal(t, 1, statusReporter.ReportInstalled[r.Recipe.Name], "Recipe1 Installed")
 	assert.Equal(t, 1, statusReporter.ReportInstalled[r2.Recipe.Name], "Recipe2 Installed")
 }
+func TestInstallGuidedShouldSkipOTEL(t *testing.T) {
+	r := &recipes.RecipeDetectionResult{
+		Recipe: recipes.NewRecipeBuilder().Name(types.InfraAgentRecipeName).Build(),
+		Status: execution.RecipeStatusTypes.AVAILABLE,
+	}
+	r2 := &recipes.RecipeDetectionResult{
+		Recipe: recipes.NewRecipeBuilder().Name(types.OTELRecipeName).Build(),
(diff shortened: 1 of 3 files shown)


internal/install/recipe_installer_test.go
```diff
@@ -168,6 +168,33 @@ func TestInstallGuidedShouldNotSkipCoreInstall(t *testing.T) {
...
+	assert.Equal(t, 1, statusReporter.RecipeDetectedCallCount, "Detection Count")
+	assert.Equal(t, 1, statusReporter.RecipeAvailableCallCount, "Available Count")
+	assert.Equal(t, 1, statusReporter.RecipeInstallingCallCount, "Installing Count")
+	assert.Equal(t, 0, statusReporter.RecipeFailedCallCount, "Failed Count")
+	assert.Equal(t, 0, statusReporter.RecipeUnsupportedCallCount, "Unsupported Count")
+	assert.Equal(t, 1, statusReporter.RecipeInstalledCallCount, "InstalledCount")
+	assert.Equal(t, 0, statusReporter.RecipeRecommendedCallCount, "Recommendation Count")
+	assert.Equal(t, 0, statusReporter.RecipeSkippedCallCount, "Skipped Count")
+	assert.Equal(t, 0, statusReporter.RecipeCanceledCallCount, "Cancelled Count")
+	assert.Equal(t, 1, statusReporter.ReportInstalled[r.Recipe.Name], "Recipe1 Installed")
```
mbruzina, line 195: Minor observation - some of these tests aren't very explicit in their failure messages. i.e. Recipe1 should install, Recipe2 should not have installed.  Might help document failures a bit more since the test specifies OTEL should be skipped (like in the target test below).  Just an observation/something we may/may not want to change.  Will not gate this PR because of it

mbruzina (approved):

Julien4218 (changes requested):
@rthorn-nr I was hoping we would add a schema element to the recipe to indicate if a recipe should be included with Guided Install. This can be defaulted to true/yes, and we would disable it for the OTel recipes. 
This way we can re-use this functionality for other recipes later on (this need came up in the past).

rthorn-nr:
Reopening to support the upcoming rescheduled OTEL targeted install release

NRhzhao (approved):

PR #1472, opened 2023-05-30
fix(install): Segment complete report and nil error when segment file missing

NRhzhao:
1. The complete count for Segment reporter reports an incorrect count since the detected status gets overwritten by other status, added a new slice to track detected and unsupported status.
2.  fix nil exception when local debugging without segment key file by returning no-op instead of nil


internal/install/command.go
```diff
@@ -104,7 +104,7 @@ func initSegment() *segment.Segment {
 	writeKey, err := recipes.NewEmbeddedRecipeFetcher().GetSegmentWriteKey()
 	if err != nil {
 		log.Debug("segment: error reading write key, cannot write to segment", err)
-		return nil
+		return segment.NewNoOp()
```
Julien4218, line 107: Good fix @NRhzhao

PR #1474, opened 2023-06-01
fix: OTEL should not be recommended for target install

NRhzhao:


rthorn-nr (approved):

stevula (approved):
Tested with nginx. Did not see OTEL recommendation and related UI crash.

PR #1475, opened 2023-06-05
refactor(release): release pipeline optimizations

sanderblue:
This PR refactors our release pipeline to have a primary release workflow that's followed by multiple secondary release flows that will run in parallel. This should dramatically speed up our overall release pipeline.

Current release pipeline duration is approximately 43 minutes.
Goal release pipeline duration is approximately 20-25 minutes.

### DIFF
--- a/.github/workflows/release-publish-windows.yml
+++ b/.github/workflows/release-publish-windows.yml
@@ -4,7 +4,7 @@ permissions: write-all
 
 on:
   workflow_run:
-    workflows: ["Release"]
+    workflows: ["Create Windows Release"]
     types: [completed]
 
 jobs:
--- a/.github/workflows/release-publish-windows.yml
+++ b/.github/workflows/release-publish-windows.yml
@@ -2,24 +2,10 @@ name: Release - Homebrew
 
 permissions: write-all
 
-# Release workflow calls this workflow
 on:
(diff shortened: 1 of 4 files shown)


NRhzhao (approved):

.github/workflows/release-publish-windows.yml
```diff
@@ -2,24 +2,10 @@ name: Release - Homebrew
...
-      AWS_ACCESS_KEY_ID:
-        required: true
-      AWS_SECRET_ACCESS_KEY:
-        required: true
-      AWS_ROLE_ARN:
-        required: true
-      AWS_DEFAULT_REGION:
-        required: true
+  workflow_run:
+    workflows: ["Release"]
```
Julien4218, line 7: @sanderblue do we need to synchronize this workflow after completion of the release-windows-installer workflow?

Julien4218 (approved):
This looks good @sanderblue , I just left a comment about workflow synchronization.

.github/workflows/release-publish-windows.yml
sanderblue: Yep, I think you're correct. Good catch. Updating now.

NRhzhao (approved):

PR #1487, opened 2023-07-18
chore: Update client get license key

NRhzhao:
Update logic on license key retrieval to have a preference
Remove license key fetching during install, should use the one retrieved during validation on earlier step


internal/install/execution/recipe_var_provider.go
```diff
@@ -77,18 +76,14 @@ func (re *RecipeVarProvider) Prepare(m types.DiscoveryManifest, r types.OpenInst
...
 	region := configAPI.GetActiveProfileString(config.Region)
 
-	if licenseKey == "" {
-		return types.RecipeVars{}, errors.New("license key not found")
-	}
-
 	vars := make(types.RecipeVars)
 
-	vars["NEW_RELIC_LICENSE_KEY"] = licenseKey
+	vars["NEW_RELIC_LICENSE_KEY"] = os.Getenv("NEW_RELIC_LICENSE_KEY")
```
mbruzina, line 86: Since we're not passing the licenseKey as a parameter any more, wondering if we should follow suit and use `configApi.GetActiveProfileString(config.LicenseKey)`, similar to the other attributes from profile

mbruzina (approved):

PR #1499, opened 2023-08-07
chore(release): attempt to fix homebrew release workflow

sanderblue:


sanderblue:
Homebrew has its own process for getting the latest releases for packages/formulae

PR #1510, opened 2023-09-05
feat(install): validate user API key format

sanderblue:
This PR introduces some light input validation to capture when customers might be using the wrong API key format and/or an invalid New Relic region during agent installation. Messaging has also been updated in hopes of providing a more helpful, descriptive error message.

### DIFF
--- a/internal/install/command.go
+++ b/internal/install/command.go
@@ -16,6 +16,7 @@ import (
 	"github.com/newrelic/newrelic-cli/internal/install/types"
 	"github.com/newrelic/newrelic-cli/internal/utils"
 	nrErrors "github.com/newrelic/newrelic-client-go/v2/pkg/errors"
+	nrRegion "github.com/newrelic/newrelic-client-go/v2/pkg/region"
 )
 
 var (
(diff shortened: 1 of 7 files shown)


internal/install/command.go
```diff
@@ -126,19 +128,31 @@ func validateProfile(maxTimeoutSeconds int, sg *segment.Segment) *types.DetailEr
...
+	if !utils.IsValidUserAPIKeyFormat(APIKey) {
+		errorOccured = true
+		detailErr = types.NewDetailError(types.EventTypes.InvalidUserAPIKeyFormat, `Invalid user API key format detected. Please provide a valid user API key. User API keys usually have a prefix of "NRAK-" or "NRAA-".`)
 		return detailErr
 	}
 
 	if region == "" {
 		errorOccured = true
-		detailErr = types.NewDetailError(types.EventTypes.RegionMissing, "region is required")
+		detailErr = types.NewDetailError(types.EventTypes.RegionMissing, "Region is required.")
```
Julien4218, line 149: @sanderblue is region really required, or do we default to US.

Julien4218 (approved):

internal/install/command.go
sanderblue: We default to the US in the client.

PR #1521, opened 2023-10-12
fix(install): remove duplicate tracking event and simplify error message

Julien4218:
Duplicate events were tracked during the validateProfile( ), this PR remove the duplicate.
Also simplified the user facing error message to avoid confusing with license key, new user message below.

`FATAL 401 response returned: Invalid credentials provided. Missing API key or an invalid API key was provided. `

### DIFF
--- a/internal/install/command.go
+++ b/internal/install/command.go
@@ -50,7 +50,6 @@ var Command = &cobra.Command{
 
 		detailErr := validateProfile(config.DefaultMaxTimeoutSeconds, sg)
 		if detailErr != nil {
-			sg.Track(detailErr.EventName)
 			log.Fatal(detailErr)
 		}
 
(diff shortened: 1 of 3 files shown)


internal/install/command.go
```diff
@@ -50,7 +50,6 @@ var Command = &cobra.Command{
 
 		detailErr := validateProfile(config.DefaultMaxTimeoutSeconds, sg)
 		if detailErr != nil {
-			sg.Track(detailErr.EventName)
```
sanderblue, line 53: Assuming we must be tracking this event on [line 124](https://github.com/newrelic/newrelic-cli/blob/fb3caec2b346fe4d7e74217af4ae23b282fe2ec8/internal/install/command.go#L124) then?

Julien4218: Exactly. I've removed that one because L124 has more context passed-in.

sanderblue (approved):

PR #1526, opened 2023-10-24
feat(entities): Support custom attributes JSON

amolero-nr:
Supports JSON strings for `--customAttributes` for deployment markers. Ref: [NR-165686](https://new-relic.atlassian.net/browse/NR-165686).

### DIFF
--- a/internal/entities/command_deployment.go
+++ b/internal/entities/command_deployment.go
@@ -3,6 +3,7 @@ package entities
 import (
 	"encoding/json"
 	"errors"
+	"fmt"
 	"log"
 	"strings"
 	"time"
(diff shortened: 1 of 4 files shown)


internal/entities/command_deployment.go
```diff
@@ -127,3 +142,23 @@ func parseCustomAttributes(a *[]string) (*map[string]string, error) {
 
 	return &customAttributeMap, nil
 }
+
+func getCustomAttributes(a []string, b string) (interface{}, error) {
```
sanderblue, line 146: Can we make this function always return `*map[string]interface{}` or `map[string]interface{}` instead of `interface{}`? Then you won't need to use the [switch](https://github.com/newrelic/newrelic-cli/pull/1526/files#diff-dddd8b36a52df433365ab87762f713ed32cb9b0282b5a2972c63608c5394bb4bR75-R80) statement.

rthorn-nr: :butting-in-emoji: `map[string]interface{}`'s zero value is nil since it's a reference type, so no need for`*map[string]interface{}`

amolero-nr: Thanks, done in ed2172c698e2995c283c80c7c47d106cb1fc9f45

PR #1534, opened 2023-11-15
feat(install): add super-agent as infra-agent substitute dependency for OHIs

amolero-nr:
Adds the ability to skip `infrastructure-agent` as a dependency for OHIs if `super-agent` is targeted. If `super-agent` is no targeted, OHIs will fall back to the original `infrastructure-agent` as their dependency.

Requires OIL changes: https://github.com/newrelic/open-install-library/pull/997

Ref: [NR-173011](https://new-relic.atlassian.net/browse/NR-173011)

### DIFF
--- a/internal/install/recipes/bundler.go
+++ b/internal/install/recipes/bundler.go
@@ -92,14 +92,7 @@ func (b *Bundler) getBundleRecipeWithDependencies(recipe *types.OpenInstallation
 	//	if super-agent is targeted, infrastructure-agent is removed
 	//	if super-agent is not targeted, super-agent is removed
 	if recipe.IsOhi() && utils.StringInSlice(types.InfraAgentRecipeName, recipe.Dependencies) && utils.StringInSlice(types.SuperAgentRecipeName, recipe.Dependencies) && depException != "" {
-		var newDeps []string
-		for _, d := range recipe.Dependencies {
-			if d == depException {
-				continue
-			}
-			newDeps = append(newDeps, d)
-		}
-		recipe.Dependencies = newDeps
+		recipe.Dependencies = utils.RemoveFromSlice(recipe.Dependencies, depException)
 	}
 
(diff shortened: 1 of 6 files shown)


internal/install/recipes/bundler.go
```diff
@@ -80,9 +88,21 @@ func (b *Bundler) getBundleRecipeWithDependencies(recipe *types.OpenInstallation
 		Recipe: recipe,
 	}
 
+	// For OHIs with dual infrastructure-agent/super-agent dependencies, remove the one not needed:
+	//	if super-agent is targeted, infrastructure-agent is removed
+	//	if super-agent is not targeted, super-agent is removed
+	if recipe.IsOhi() && utils.StringInSlice(types.InfraAgentRecipeName, recipe.Dependencies) && utils.StringInSlice(types.SuperAgentRecipeName, recipe.Dependencies) && depException != "" {
```
NRhzhao, line 94: We don't want to hardcode logic based on recipe name (such as super-agent).
We just want to check if there is a target install recipe:
If there is no target install: add/check the first dependency in order.
If there is a target: check if the target fulfills the dependency requirement. If it doesn't, add/check the first dependency in order. If it does, then we don't need to add any dependency, just install the target since it already targeting a dependency.   There might be some additional logic modification downstream where if install failed from the target dependency, we want to not force install the other dependency, and fail to install the OHI.

NRhzhao, line 94: I don't think we need to have this IsOhi() check, this behavior can be applicable to all recipes.

amolero-nr: To discuss, a new PR: https://github.com/newrelic/newrelic-cli/pull/1536

amolero-nr: That check was removed in this [new PR](https://github.com/newrelic/newrelic-cli/pull/1536).

PR #1562, opened 2024-02-28
chore: Measure ingest latency

rthorn-nr:
This PR adds segment tracking to measure latency during the install step "Connecting to New Relic Platform"


internal/diagnose/segment.go
```diff
@@ -0,0 +1,27 @@
...
+import (
+	"github.com/newrelic/newrelic-cli/internal/config"
+	configAPI "github.com/newrelic/newrelic-cli/internal/config/api"
+	"github.com/newrelic/newrelic-cli/internal/install/recipes"
+	"github.com/newrelic/newrelic-cli/internal/segment"
+	log "github.com/sirupsen/logrus"
+	"golang.org/x/net/http/httpproxy"
+)
+
+func initSegment() *segment.Segment {
```
amolero-nr, line 12: It's nice that we now have a `segment` package, I wonder if we could consolidate the `initSegment` (and) functions inside `internal/diagnose/segment.go` and `internal/install/command.go` into one inside that new package to not duplicate that code...

A similar thought possibly for `IsProxyConfigured` too, which we have in `internal/install/proxy_configuration.go` and `internal/diagnose/segment.go`.

internal/install/recipe_installer.go
```diff
@@ -271,15 +273,25 @@ func (i *RecipeInstall) connectToPlatform() error {
...
 	}()
 
+	i.segment.Track("ConnectingToPlatformStarted")
 	i.progressIndicator.Start("Connecting to New Relic Platform")
+	start := time.Now()
 
 	loaded := <-loaderChan
 
+	durationMs := time.Since(start).Milliseconds()
+	ei := segment.NewEventInfo("ConnectingToPlatformComplete", "")
```
amolero-nr, line 283: A follow up item in a separate Jira could be adding/defining the new events (`ConnectingToPlatform*` and `ValidateNrql*`) to the `internal/install/types/errors.go` and even assessing, in case those events are only related to segment tracking, to extract those into a `types.go` file inside the new `segment` package you refactored.

internal/diagnose/segment.go
rthorn-nr: Good call, moved these over to `segment`. This caused an import cycle at first, but I resolved that by extracting the write key read out of `install/recipes` and over to another embedded FS in `segment`

internal/install/recipe_installer.go
```diff
@@ -271,15 +273,25 @@ func (i *RecipeInstall) connectToPlatform() error {
 		loaderChan <- nil
 	}()
 
+	i.segment.Track("ConnectingToPlatformStarted")
```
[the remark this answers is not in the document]
rthorn-nr: In the spirit of trying to cut down on scope pollution for those structs, I was able to move them out of both `ConfigValidator` and `RecipeInstall` and into a package-scoped variable. This way we can create events anywhere without changes to structs or drilling down

internal/install/recipe_installer.go
rthorn-nr: For sure, would be great to define these alongside the other events. It seemed simplest to me to set these up outside of the reporting system. There's probably more cleanup to do around segment logic built inside `install/recipes`. 

If it's not in a good state to you, I can address any of this in this PR, otherwise a follow-up PR could include a closer look at extracting everything else we can into `segment` and the event type cleanup you're suggesting

amolero-nr: I think following up in a separate ticket would be best for now.

Issue #1292, opened 2022-05-06
Not possible to search for some entity types


abuchanan-nr:
### Description
It's not possible to search for some valid entity types, such as AIOPS/CONDITION.

### Go Version
go version go1.16.5 darwin/amd64
newrelic version 0.47.0

### Current behavior
```
> newrelic entity search "accountId = 1234 and type = 'CONDITION'"
FATAL one of --name, --type, --alert-severity, or --domain are required

> newrelic entity search --domain 'AIOPS' "accountId = 1234 and type = 'CONDITION'"
FATAL Argument "queryBuilder" has invalid value $queryBuilder.
In field "domain": Expected type "EntitySearchQueryBuilderDomain", found "AIOPS".

> newrelic entity search --type CONDITION "accountId = 1234 and type = 'CONDITION'"
FATAL Argument "queryBuilder" has invalid value $queryBuilder.
In field "type": Expected type "EntitySearchQueryBuilderType", found "CONDITION".
```

### Expected behavior
It should be possible to search for entity types that are not part of `EntitySearchQueryBuilderType`. 

It should be possible to search without using the `entitySearch.queryBuilder`, but instead use `entitySearch.query`:
```
{
  actor {
    entitySearch(query: "(tags.owning_team = 'abcd' and type != 'CONTAINER') or (accountId = 1234)") {
      results {
        entities {
          name
          type
          tags {
            key
            values
          }
        }
      }
    }
  }
}
```

maxlemieux:
This is also affecting searches for OpenTelemetry service entities. Using this query does not work as expected:

```
newrelic entity search --domain=EXT --type=SERVICE
FATAL Argument "queryBuilder" has invalid value $queryBuilder.
In field "type": Expected type "EntitySearchQueryBuilderType", found "SERVICE". 
```

Reference for entity domain and type: https://github.com/newrelic/entity-definitions/blob/main/entity-types/ext-service/definition.yml

Issue #138, opened 2020-03-13
Enable non-fuzzy Entity Searching


jthurman42:
### Description
The current implementation of entity search returns a set of fuzzy matched entities, which can make it difficult when you are trying to filter to a specific name.

### Acceptance Criteria
- Able to search on strict values (i.e.  `--name "webportal"` should hard-match)
- Ability to fuzzy search (Suggestion: use NRQL syntax for fuzzy searching: `--name "%webportal%"`)

jthurman42:
User testing feedback actually preferred the opposite.  Default to fuzzy (like most cli tools do), but would be nice to enable some non-fuzzy syntax.  Unsure how we could do that since downstream is all fuzzy matching (without re-implementing our own search queries)

lhenry7:
A name search would imply an exact match to me, or at least deserves the option to make it so.

PR #1570, opened 2024-03-15
feat(entities): upgrade entity search to support more entity types and options

sanderblue:
Requires: [newrelic/newrelic-client-go #1113](https://github.com/newrelic/newrelic-client-go/pull/1113)

Fixes: #1292 
Fixes: #138

### DIFF
--- a/internal/entities/command_search.go
+++ b/internal/entities/command_search.go
@@ -1,7 +1,7 @@
 package entities
 
 import (
-	"strconv"
+	"context"
 
 	log "github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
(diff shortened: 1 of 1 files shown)


internal/entities/command_search.go
```diff
@@ -12,61 +12,44 @@ import (
...
+		query := entities.BuildEntitySearchNrqlQuery(searchParams)
 
-		results, err := client.NRClient.Entities.GetEntitySearchWithContext(
-			utils.SignalCtx,
-			entities.EntitySearchOptions{},
-			"",
-			params,
+		results, err := client.NRClient.Entities.GetEntitySearchByQueryWithContext(context.Background(),
+			entities.EntitySearchOptions{
+				CaseSensitiveTagMatching: entitySearchCaseSensitive,
```
pranav-new-relic, line 50: A thought I'd like to share (not for now, maybe sometime in future for one of us to add) - `entities.EntitySearchOptions` also has a `limit` field that helps specify the number of entities to be returned in results - do you think it would be a good idea for us to facilitate this via a new flag in the `newrelic entity search` command

internal/entities/command_search.go
```diff
@@ -107,11 +90,13 @@ func mapEntities(entities []entities.EntityOutlineInterface, fields []string, fn
...
-	cmdEntitySearch.Flags().StringVarP(&entityDomain, "domain", "d", "", "search for entities matching the given entity domain")
-	cmdEntitySearch.Flags().StringVar(&entityTag, "tag", "", "search for entities matching the given entity tag")
-	cmdEntitySearch.Flags().StringSliceVarP(&entityFields, "fields-filter", "f", []string{}, "filter search results to only return certain fields for each search result")
+	cmdEntitySearch.Flags().StringVarP(&entityName, "name", "n", "", "Search for entities matching the given name")
+	cmdEntitySearch.Flags().StringVarP(&entityType, "type", "t", "", "Search for entities matching the given type")
+	cmdEntitySearch.Flags().StringVarP(&entityDomain, "domain", "d", "", "Search for entities matching the given entity domain")
+	cmdEntitySearch.Flags().StringVarP(&entityAlertSeverity, "alert-severity", "s", "", "Search for entities matching the given alert severity type")
+	cmdEntitySearch.Flags().StringVarP(&entityReporting, "reporting", "r", "", "Search for entities based on whether or not an entity is reporting (true or false)")
+	cmdEntitySearch.Flags().StringVar(&entityTag, "tag", "", "Search for entities matching the given entity tag")
```
pranav-new-relic, line 98: I think we no longer need the flag `tag` since we have `tags`, also because we aren't really using the flag `tag` anywhere in the implementation of this command. Thoughts?

pranav-new-relic (approved):

NSSPKrishna (dismissed):

internal/entities/command_search.go
sanderblue: Good catch, however, since `--tag` was a defined flag previously, I don't want to delete it since that would be a breaking change. 

Also I appreciate you catching the `reporting` part. I missed including that functionality in this refactor. That would also be a breaking change. I'll push up a new commit for that shortly.

pranav-new-relic (approved):

PR #1604, opened 2024-05-09
feat(super-agent): Skip core bundle if installed

NSSPKrishna:
This pull request uses the CLI functionality to detect the presence of the Super Agent on the host. If detected, it will install OHI with the Super Agent dependency, skipping the core bundle installation. If core recipes are targeted for installation, they will be unsupported since the Super Agent is already installed. (Associated PR: https://github.com/newrelic/open-install-library/pull/1072)

### DIFF
--- a/internal/install/bundle_installer.go
+++ b/internal/install/bundle_installer.go
@@ -3,6 +3,7 @@ package install
 import (
 	"context"
 	"fmt"
+	"github.com/newrelic/newrelic-cli/internal/utils"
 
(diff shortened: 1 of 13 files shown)


NSSPKrishna:
Updated the status of the core bundle to NULL from UNSUPPORTED when a super agent is installed, as suggested by @sblue. This modification enables easier tracking of installs, particularly in scenarios where a super agent is present.

Avinash-Sanpala (approved):
LGTM!

PR #1609, opened 2024-05-20
fix(infra): Remove Test endpoints to fix issue with infra-alert.service

Nandu-pns:
JIRA - https://new-relic.atlassian.net/browse/NR-270753 

This PR aims to address the below timeout issue for few customers as it appears CLI is performing few [test calls ](https://github.com/newrelic/newrelic-cli/blob/53c2d81b257fe874d86c4a0b97b14489d9e7cd99/internal/install/command.go#L178) to verify if the New Relic endpoints are reachable and this infra-alert.service is being invoked during this specific [call](https://github.com/newrelic/newrelic-client-go/blob/0e073a6f00c54832713063f6e37222759337ff41/pkg/region/region_constants.go#L27).

`FATAL Get "https://login.newrelic.com/login?return_to=https%3A%2F%2Finfrastructure-alert.service.newrelic.com%2Fv2": dial tcp 18.188.220.94:443: i/o timeout`

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -16,7 +16,7 @@ require (
 	github.com/jedib0t/go-pretty/v6 v6.4.4
 	github.com/joshdk/go-junit v0.0.0-20210226021600-6145f504ca0d
 	github.com/mitchellh/go-homedir v1.1.0
-	github.com/newrelic/newrelic-client-go/v2 v2.36.2-0.20240527091533-de75a293dda1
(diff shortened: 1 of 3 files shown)


Nandu-pns:
Closing this PR as we plan to discard the specific infra test API call which is causing the issue instead of removing `TestEndpoints` function call. Changes are in the client-go [PR](https://github.com/newrelic/newrelic-client-go/pull/1145/files)

PR #1621, opened 2024-06-14
feat: Display unsupported errors for targeted recipes

nmcgi:
Displays errors when checking requirements at discovery, including processes and dependencies, ONLY if the user has targeted the recipe.

This will provide more context around "Unsupported" errors when attempting to install a specific install.

Guided install will continue to suppress "Unsupported" error messaging.


sanderblue (approved):
Nice :)

PR #1628, opened 2024-07-10
chore: Added support for reading quey from a file

jsumners-nr:
Prior to this change, trying to do the following will result in an error:

```sh
$ newrelic nrql query --query <(cat some.nrql)
```

With this change, we support proper shell redirection. And, technically, the following will also work:

```sh
$ newrelic nrql query --query ./some.nrql
```


internal/nrql/command_query.go
```diff
@@ -29,6 +32,19 @@ issue the query against.
...
+		} else {
+			// query points to a file/file descriptor
+			fileBytes, readErr := os.ReadFile(query)
+			if readErr != nil {
+				log.Fatal(readErr)
+			}
+			query = string(fileBytes)
+		}
+
 		result, err := client.NRClient.Nrdb.QueryWithContext(utils.SignalCtx, accountID, nrdb.NRQL(query))
```
jsumners-nr, line 48: It'd be nicer if `QueryWithContext` accepted an `io.Reader` instead of a string. But that seems like a more difficult change.

PR #1637, opened 2024-08-05
Fix: Remove AB testing

noahmmcgivern:
Removes A/B testing implementation that utilized Split.io and Segment.io:
- Removes Segment.io client implementation and tests
- Removes Segment Reporter implementation and tests
- Removes Split.io server implementation and tests
- Removes references to Segment.io/Split.io in various functions
- Removes respective direct and indirect dependencies
- Removes reference to keys
- Removes events Makefile and its references
- Removes A/B section in documentation


sanderblue (approved):
🧹 🧹

PR #1645, opened 2024-09-17
fix: Uninstall previous cli version windows

noahmmcgivern:
Modifies the install.ps1 script to check for existing instances of the CLI. Removes all existing instances of the CLI before proceeding to the (re)installation.


sanderblue (approved):
🎉

PR #1669, opened 2024-11-22
fix(dashboard): Updated the Dashboard JSON to HCL Converter Utility with Latest Fields in the Resource

vinay-newrelic:
Problem Statement:

The problem at hand is to update the CLI dashboards utility which helps convert JSON dashboards into HCL-based dashboards (generating the newrelic_one_dashboard resource HCL configuration, from the JSON version of a dashboard) with the latest fields added to the Terraform resource, as these relatively new fields are unsupported by the CLI tool, owing to their absence in the parser of the tool, that needs an update

Solution :

schema of this CLI utility is updated with the latest fields updated in the newrelic_one_dashboard resource

### DIFF
--- a/internal/utils/terraform/dashboard.go
+++ b/internal/utils/terraform/dashboard.go
@@ -2,6 +2,7 @@ package terraform
 
 import (
 	"encoding/json"
+	"fmt"
 	"strings"
 
 	log "github.com/sirupsen/logrus"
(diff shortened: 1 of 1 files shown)


pranav-new-relic (approved):
Nice work :)

You'll need to release the CLI to have these changes published. The release process is similar to that of the Terraform Provider 👍

Issue #1568, opened 2024-03-13
No output when creating muting rules


ghost:
### Description
When I create a muting rule with newrelic-cli, the rule is properly created but output JSON is empty.
Muting rule Id should be returned as it is with curl command.

### Go Version
Go not installed, use of newrelic-cli
```
$ newrelic --version
newrelic version 0.78.17
```

### Current behavior
When I run this command (generated from NerdGraph API Explorer), returned JSON is empty
```
NEW_RELIC_REGION=eu NEW_RELIC_API_KEY=XXXXX newrelic nerdgraph query 'mutation {
  alertsMutingRuleCreate(
    accountId: 00000
    rule: {name: "dev-paris deployment 2", description: "Muting rule created with newrelic-cli", enabled: true, condition: {conditions: {attribute: "conditionId", operator: IN, values: ["XXXX", "YYYY"]}, operator: AND}}
  ) {
    id
  }
}' 
{}
```

### Expected behavior
Get the same output than curl command:
```
NEW_RELIC_REGION=eu NEW_RELIC_API_KEY=XXXXX newrelic nerdgraph query 'mutation {
  alertsMutingRuleCreate(
    accountId: 000000
    rule: {name: "dev-paris deployment 2", description: "Muting rule created with newrelic-cli", enabled: true, condition: {conditions: {attribute: "conditionId", operator: IN, values: ["XXXX", "YYYY"]}, operator: AND}}
  ) {
    id
  }
}' 
{"data":{"alertsMutingRuleCreate":{"id":"XXXXX"}}}
```

### Steps To Reproduce
1. Create `alertsMutingRuleCreate` query in nergraph API explorer
2. Copy as newrelic-cli
3. Run command

PR #1670, opened 2024-12-04
feat(nerdgraph): Add support to output results for mutations in CLI

vinay-newrelic:
https://github.com/newrelic/newrelic-cli/issues/1568
image

### DIFF
--- a/internal/nerdgraph/command_query.go
+++ b/internal/nerdgraph/command_query.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"encoding/json"
 	"errors"
+	"github.com/newrelic/newrelic-cli/internal/output"
 	"io/ioutil"
 	"strings"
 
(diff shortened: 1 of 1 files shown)


Nandu-pns:
@vinay-newrelic If possible, Could you update the PR description with the CLI command outputs after the changes? Thanks !

internal/nerdgraph/command_query.go
```diff
@@ -4,16 +4,15 @@ import (
...
 	"io/ioutil"
 	"strings"
 
 	log "github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
 
 	"github.com/newrelic/newrelic-cli/internal/client"
-	"github.com/newrelic/newrelic-cli/internal/output"
 	"github.com/newrelic/newrelic-cli/internal/utils"
-	ng "github.com/newrelic/newrelic-client-go/v2/pkg/nerdgraph"
```
shashank-reddy-nr, line 16: why is this line removed?

vinay-newrelic: Since there is no longer a need of converting the response into particular nerdgraph fields (Actor,docs,reqestcontext) which exists only in the case of query response and not mutation response.

shashank-reddy-nr: Makes sense

shashank-reddy-nr (approved):
LGTM

PR #1677, opened 2025-01-27
fix(dependabot): Updated dependencies to fix dependabot alerts

rpaliwal1997:
[Jira](https://new-relic.atlassian.net/browse/NR-360591)

Dependabot Alerts issues:-

[Issue 1](https://github.com/newrelic/newrelic-cli/security/dependabot/64)
[Issue 2](https://github.com/newrelic/newrelic-cli/security/dependabot/66)
[Issue 3](https://github.com/newrelic/newrelic-cli/security/dependabot/63)
[Issue 4](https://github.com/newrelic/newrelic-cli/security/dependabot/67)
[Issue 5](https://github.com/newrelic/newrelic-cli/security/dependabot/57)
[Issue 6](https://github.com/newrelic/newrelic-cli/security/dependabot/65)
[Issue 7](https://github.com/newrelic/newrelic-cli/security/dependabot/52)
[Issue 8](https://github.com/newrelic/newrelic-cli/security/dependabot/49) 
[Issue 9](https://github.com/newrelic/newrelic-cli/security/dependabot/48)
[Issue 10](https://github.com/newrelic/newrelic-cli/security/dependabot/62)

I updated the dependencies for above issues to avoid vulnerabilities.

PR #1680, opened 2025-02-04
fix: Revert PR #1677 causing regression test failure

shashank-reddy-nr:
### Changes:
- Reverted PR#1677


pranav-new-relic (approved):
Looks good to me, thank you for spotting this @shashank-reddy-nr.

@pranav-new-relic and @rpaliwal1997 to look into bringing back changes in the dependabot PR, making sure these failures no longer occur

PR #1694, opened 2025-03-13
fix: Bumped the golangci-lint version

shashank-reddy-nr:


shashank-reddy-nr:
Clone branch isn't properly bumped the indirect dependency packages. Due to this we are seeing lint issue. Informed @rpaliwal1997 to update the direct dependency packages always. We shouldn't directly update indirect dependency packages. Suggested the correct procedure.

PR #1707, opened 2025-04-21
fix: Updated the ubuntu version in rel ymls

shashank-reddy-nr:
### Changes:
- Updated the ubuntu versions in release ymls. This is due to current version ubuntu-20.04 is recently deprecated and no longer supported. Followed below solution to fix it.

**Suggested solutions: https://github.com/actions/runner-images/issues/11101**

#### Error:
image

### DIFF
--- a/.github/workflows/release-publish-windows.yml
+++ b/.github/workflows/release-publish-windows.yml
@@ -23,7 +23,7 @@ on:
 jobs:
   release-windows:
     name: Publish Windows Release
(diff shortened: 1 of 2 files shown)


vinay-newrelic (approved):

Nandu-pns (commented):
Changes looks good to me !

Nandu-pns (approved):

PR #1718, opened 2025-05-26
fix(dependabot-alerts): bump github.com/goreleaser/goreleaser to v1.26.2

Aashirwadjain:
### Description

This PR aims to migrate goreleaser to v1.26.2

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -41,7 +41,14 @@ archives:
     format_overrides:
       - goos: windows
         format: zip
-    name_template: "{{ .ProjectName }}_{{ .Version }}_{{ title .Os }}_{{ .Arch }}"
+    name_template: >-
+      {{- .ProjectName }}_
+      {{- .Version }}_
+      {{- title .Os }}_
+      {{- if eq .Arch "amd64" }}x86_64
+      {{- else if eq .Arch "386" }}i386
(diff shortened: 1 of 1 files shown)


.goreleaser.yml
```diff
@@ -169,15 +171,16 @@ scoop:
 nfpms:
   - id: newrelic
     package_name: newrelic-cli
-    file_name_template: "{{ .ProjectName }}_{{ .Version }}_{{ .Os }}_{{ .Arch }}"
+    file_name_template: >-
+      {{- .ProjectName }}_
+      {{- .Version }}_
+      {{- title .Os }}_
+      {{- if eq .Arch "amd64" }}x86_64
```
shashank-reddy-nr, line 178: Can you please check if we can do this without using if else conditions. In yaml we generally avoid using if conditions.

Aashirwadjain: Hi @shashank-reddy-nr, This approach is used particularly to avoid using replacements block. For this I have used `file_name_template`, more information on this can found [here](https://goreleaser.com/deprecations/?h=replacements#nfpmsreplacements).

This approach is mentioned in [goreleaser example repositories](https://github.com/goreleaser/example-deno/blob/main/.goreleaser.yaml#L25). However, I tried to find other approaches which makes use of ternary operator or custom functions, but found this approach to be most effective.

PR #1720, opened 2025-05-26
fix: Add InstallEventsReporter only if API key is set

rpaliwal1997:
[Jira Ticket](https://new-relic.atlassian.net/browse/NR-377103)

**Problem:-**

Customers were encountering the following error during execution:

`Debug Error: 401 response returned: Invalid credentials provided. Missing API key or an invalid API key was provided.`

**Cause:-**

The 401 Unauthorized error occurred because the recipe attempted to authenticate with New Relic's API using a License Key, which is not valid for API authentication. API operations require a valid API Key (User API Key or Personal API Key).

**Solution:-**

This fix updates the code to only add the **InstallEventsReporter status subscriber** if a valid API Key is present in the active profile. This prevents attempts to authenticate with invalid credentials and avoids unnecessary 401 errors.

**Testing Results:-** No  401 Unauthorized error after the fix.

image

image

### DIFF
--- a/internal/install/recipe_installer.go
+++ b/internal/install/recipe_installer.go
@@ -73,10 +73,18 @@ func NewRecipeInstaller(ic types.InstallerContext) *RecipeInstall {
 	mv := discovery.NewManifestValidator()
 	ff := recipes.NewRecipeFileFetcher([]string{})
 	lf := execution.NewRecipeLogForwarder()
+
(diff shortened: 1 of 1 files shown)


shashank-reddy-nr:
Nice work @rpaliwal1997 , changes looks good.

shashank-reddy-nr (approved):

PR #1721, opened 2025-05-29
feat: Added support for using agent rel versions in CLI install cmds

shashank-reddy-nr:
Ticket: https://new-relic.atlassian.net/browse/NR-419096

### Changes:
- Support for specifying agent/recipe release versions in CLI commands has been added. This feature will be useful for customers who wish to install a particular version of the agent. Currently, the CLI does not support this functionality, and we are always installing the latest version.

### Note:
The Authers(the agent teams ex: infra, k8s, OHIs, APMs) have to read version variable from env. This feature is setting version to env with recipe name followed by _version. So, the support for this has to be added to recipes.

Ex: lets take k8s recipe, the recipe name is `kubernetes-open-source-integration`. So, we are setting KUBERNETES_OPEN_SOURCE_INTEGRATION_VERSION to env variables with the values user provided in command. if not provided the latest has to be installed. this should be the ideal soltuin with this change.

I will soon create a few examples on how authors can do for their respective recipes, if needed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,7 @@
+<a name="v0.98.4"></a>
+## [v0.98.4] - 2025-05-27
 <a name="v0.98.3"></a>
-## [v0.98.3] - 2025-05-26
+## [v0.98.3] - 2025-05-27
 ### Bug Fixes
 - Add InstallEventsReporter only if API key is set ([#1720](https://github.com/newrelic/newrelic-cli/issues/1720))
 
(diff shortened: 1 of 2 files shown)


internal/install/command.go
```diff
@@ -27,18 +27,54 @@ var (
...
 
+// processRecipeNames validates, extracts recipe names, and sets environment variables.
+func processRecipeNames(recipeNames []string) ([]string, error) {
+	var extractedNames []string
+
+	for _, recipe := range recipeNames {
+		parts := strings.Split(recipe, "@")
+
+		if len(parts) < 1 || len(parts) > 2 {
+			return nil, fmt.Errorf("invalid recipe name format provided: %s", recipe)
```
vagrawal-newrelic, line 38: @shashank-reddy-nr Just a small update if it is possible to let the user know what is the correct format along with the error message will be helpful, may be a link to the docs that will have example of this support.

shashank-reddy-nr: Sure, Noted. Once, NR docs gets updated. I will add docs link here.

vinay-newrelic (approved):

shashank-reddy-nr:
Ubuntu AMI image is deprecated, So, test is failing. We will update it in coming sprint.

image

PR #1725, opened 2025-06-25
fix(deps): update dependencies in response to a few vulns reported

pranav-new-relic:


pranav-new-relic:
Merging this PR, as the only failing test is owing to an AMI mistmatch which should be unrelated to these changes in deps (which I've verified with a bunch of other PRs too

nr-developer-toolkit (approved):

PR #1731, opened 2025-07-16
fix(checksum): Add install.sh and install.ps1 for checksum generation

vagrawal-newrelic:
fix(checksum): Add install.sh and install.ps1 for checksum generation


pranav-new-relic (approved):

pranav-new-relic:
🤞

PR #1741, opened 2025-08-04
fix: Skip lsass.exe process from being processed

shashank-reddy-nr:
### Changes:
- Skipped LSASS.exe process to be processed for processMatch.

### Test Results (with debug prints):
image

### Before Change:
image

### No LSASS.exe in MDE timeline logs Now:
image


NSSPKrishna (approved):
LGTM

PR #1736, opened 2025-07-17
fix(checksum): Remove the Manuel install.ps1 upload

vagrawal-newrelic:
This is the final PR  that include changes to upload the install.sh and install.ps1 files to the GITHUB assets directory and also generating and adding the checksum to the checksum file for validating the install.sh file at customer end, As part of this [WR](https://new-relic.atlassian.net/browse/NR-307933)

To fulfil the requirement below PR include the consecutive changes
[1731](https://github.com/newrelic/newrelic-cli/pull/1731):   Include changes to upload install.sh and install.ps1 to GIt assets using goreleaser config
[1734](https://github.com/newrelic/newrelic-cli/pull/1734): include changes to add the checksum for the install.sh and install.ps1 in the checksum file

This PR include removal of redundant/Manuel step of uploading the install.ps1 to git assets  as this is now covered as part of the above PR changes.

Test:

image

image

PR #1743, opened 2025-08-04
feat(checksum): Update checksum signing config

vagrawal-newrelic:
This is a PR that include minor update where we are adding a new step in the release.yml file to copy the checksum file with a different name and push it to GIT release assets along with its respective signature file.
This changes are done for the minor request that was asked by the customer(E.ON) in this [WR](https://new-relic.atlassian.net/browse/NR-307933).

Changes for the checksum generation and adding the install scripts to the GIT release assets was done as part of https://github.com/newrelic/newrelic-cli/pull/1736 PR.

No Test validation SS is available as there is no way to test the changes in local, a realtime release is required to test the changes..

Any further changes on this will be part of other PR and this PR will be attached in the new one to keep a track of the PR's that include all relevant changes to cater this WR.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -180,7 +180,10 @@ jobs:
           echo "Created checksums.txt successfully"
           
           # Create signature and validate
-          gpg --detach-sign --armor -u "4F9A9B5B96EC30B9" "dist/checksums.txt"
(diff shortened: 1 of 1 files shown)


shashank-reddy-nr (approved):
LGTM

PR #1748, opened 2025-09-03
fix(agent-control): use proper fleet link

sigilioso:
This PR fixes the link shown when the Agent Control is successfully installed in a host. The link was resolving to `<domain>/fleet)` which was redirecting to the `nr1-core` main page. The fixed link will point to the list of Fleets.


NSSPKrishna (approved):
LGTM, great catch! 😄

PR #1729, opened 2025-07-11
feat(change-event): adds support for the change tracking api

NSSPKrishna:
Description: 
This is for the change tracking. 

Testing:

```bash
make compile-only
./bin/darwin/newrelic changeTracking --help
```

PR #1752, opened 2025-10-13
chore: Add docs page for changeTracking command

kaacarroll:
The `changeTracking` command was added in https://github.com/newrelic/newrelic-cli/pull/1729 but the docs readme was not added in this directory, so just adding it for good measure.


pranav-new-relic:
@Nandu-pns can you please take a look at this tomorrow and approve it as well (just so we're sure) so we can have it merged? Thanks!

https://newrelic.slack.com/archives/C01PTDC51K2/p1761065941724079

PR #1840, opened 2026-05-06
fix(autodiscovery): add `NEW_RELIC_SKIP_AUTODISCOVERY` env var to skip autodiscovery for targeted installs

pranav-new-relic:
## Summary

When running a targeted install (e.g. `newrelic install -n agent-control`), the CLI evaluates `requireAtDiscovery` scripts for all known recipes — currently 87+ — before building the install bundle. This includes expensive checks like the MSSQL registry scan (~2 minutes) even when the user only asked for a single specific recipe.

Setting `NEW_RELIC_SKIP_AUTODISCOVERY=1` (or any truthy value: `true`, `TRUE`, `t`) alongside a `-n` flag now limits discovery to only the explicitly requested recipes. All others immediately return a `NULL` status, which is a no-op in the reporting pipeline, so no existing output or event reporting is affected.

The guard requires both the env var **and** a targeted recipe (`-n` or `-c`). Guided installs — where no recipe is named — are entirely unaffected even if the env var is set, preserving the existing discovery behaviour for interactive users.

`strconv.ParseBool` is used for the truthy check rather than a raw `== "1"` comparison. This was necessary after field testing revealed that PowerShell users naturally write `$env:NEW_RELIC_SKIP_AUTODISCOVERY='true'`, which the original `== "1"` check silently ignored.

Documentation has been added in three places: the `Long` description on the cobra command (visible via `newrelic install --help`), `docs/cli/newrelic_install.md`, and the env var reference table in `docs/GETTING_STARTED.md`.

## Note

The commit `3c1c64c2` includes a `newrelic` binary that was temporarily committed to share a test build with the Agent Control team. This should be removed before merge — either by amending the commit or dropping the file in a follow-up squash.

## Test plan

- [ ] `go test ./internal/install/recipes/... -run "TestRecipeDetector|TestSkipAutodiscovery"` — 13 tests covering the new env var behaviour, truthy value forms, guided install safety, and unchanged default behaviour
- [ ] Manually verified by Agent Control team (Victor Ripoll) against a Windows host running `newrelic install -n agent-control --debug` with `$env:NEW_RELIC_SKIP_AUTODISCOVERY=1` — autodiscovery logs for non-targeted recipes no longer appear

### DIFF
--- a/docs/GETTING_STARTED.md
+++ b/docs/GETTING_STARTED.md
@@ -246,6 +246,7 @@ Custom events may take a moment to appear in query results while they are being 
 | `NEW_RELIC_API_KEY`                    | Your New Relic [User API key] \(prefixed with `NRAK`).         |
 | `NEW_RELIC_REGION`                     | Your New Relic account's [data center region] \(`US`, `EU`, or `JP`). |
 | `NEW_RELIC_CLI_SKIP_CORE`              | Skipping core recipes during installation \(`1` or `0`).       |
+| `NEW_RELIC_SKIP_AUTODISCOVERY`         | When combined with `-n`, skip auto-discovery for all recipes not explicitly requested. Reduces installation time in automated environments by preventing `requireAtDiscovery` scripts from running for every known recipe. Has no effect without `-n`. Accepts truthy values: `1`, `true`, `TRUE`. |
 
 ### Want more?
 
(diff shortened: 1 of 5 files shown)


shashank-reddy-nr (approved):
LGTM

PR #1841, opened 2026-05-08
fix(ci): add explicit permissions to GitHub Actions workflows

Nandu-pns:
Resolves NR-560197 — CodeQL flagged 17 `actions/missing-workflow-permissions` alerts across 11 workflow files.

## Summary

- Added least-privilege `permissions:` blocks at the workflow level for all 11 affected files                                                                                                                                                                                            
- 10 workflows need read-only access: `contents: read`                                                                                                                                                         
- `release-oil-then-cli.yml` also needs `actions: write` to trigger and list workflow runs via `gh workflow run` / `gh run list` using `GITHUB_TOKEN`

### DIFF
--- a/.github/workflows/cleanup-eu.yml
+++ b/.github/workflows/cleanup-eu.yml
@@ -5,6 +5,9 @@ on:
     workflows: [End to end testing (EU), Release]
     types: [completed]
 
+permissions:
+  contents: read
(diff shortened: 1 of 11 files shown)


pranav-new-relic (approved):
LGTM

PR #1842, opened 2026-05-08
feat(iam): add usermanagement and accessmanagement CLI commands

jt-nr:
Adds two new top-level commands that expose New Relic user and access management via the CLI.

## Command structure

The command hierarchy mirrors the two distinct sections of the New Relic Administration UI — User Management and Access Management — rather than the underlying SDK package layout. This gives CLI users a mental model consistent with the UI and the product documentation, regardless of how the NerdGraph schema is organized internally.

`usermanagement` (actor.organization.userManagement.*):
- users get/create/update/delete
- groups get/create/update/delete
- groups members add/remove
- auth-domains get

`accessmanagement`:
- grants get/create/revoke
- roles get
- permissions get

## SDK usage in accessmanagement

The Access Management UI makes NerdGraph calls to two different schema roots on the same page: actor.organization.authorizationManagement.* for grant mutations, and customerAdministration.* for reads (grants get, roles get, permissions get). We followed the same split here so CLI behavior matches what users see in the UI:

- grants create/revoke → authorizationManagement SDK
- grants get, roles get, permissions get → customerAdministration SDK

roles get uses customerAdministration specifically because it supports richer filtering (name contains, groupId) compared to the equivalent authorizationManagement method.

## Naming

- "grants create/revoke" follows the CRUD convention used elsewhere in the CLI rather than the SDK's GrantAccess/RevokeAccess naming.
- Flag names use camelCase to match existing CLI conventions (--authDomainId, --groupId, --userId, --roleId).

Includes per-package READMEs with command reference, flag docs, example responses, jq recipes, and an end-to-end workflow that spans both commands — since the natural flow is to create users/groups in usermanagement and then assign access in accessmanagement.

### DIFF
--- a/cmd/newrelic/main.go
+++ b/cmd/newrelic/main.go
@@ -9,6 +9,7 @@ import (
 	flag "github.com/spf13/pflag"
 
 	// Commands
+	"github.com/newrelic/newrelic-cli/internal/accessmanagement"
 	"github.com/newrelic/newrelic-cli/internal/agent"
 	"github.com/newrelic/newrelic-cli/internal/apiaccess"
 	"github.com/newrelic/newrelic-cli/internal/apm"
@@ -31,6 +32,7 @@ import (
 	"github.com/newrelic/newrelic-cli/internal/profile"
 	"github.com/newrelic/newrelic-cli/internal/reporting"
 	"github.com/newrelic/newrelic-cli/internal/synthetics"
+	"github.com/newrelic/newrelic-cli/internal/usermanagement"
 	"github.com/newrelic/newrelic-cli/internal/utils"
 	"github.com/newrelic/newrelic-cli/internal/workload"
 )
@@ -41,6 +43,7 @@ var (
 
 func init() {
 	// Bind imported sub-commands
+	Command.AddCommand(accessmanagement.Command)
 	Command.AddCommand(agent.Command)
 	Command.AddCommand(apiaccess.Command)
 	Command.AddCommand(synthetics.Command)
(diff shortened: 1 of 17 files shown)


jt-nr:
Note on the E2E failures: these are expected for fork PRs. The `e2e.yml` workflow requires repository secrets (New Relic API keys, account IDs, cloud credentials) to provision agents on real infrastructure — GitHub doesn't pass those to workflows triggered from forks. No code changes needed. A maintainer will need to trigger E2E after review.

jt-nr:
Rebased onto current `main` (v0.113.19) to clear the stale branch state. The branch was 91 commits behind. The rebase was conflict-free since every file in this PR is a new file, and the diff is unchanged at +1846/-0 across 17 files.

All non-E2E checks are green on the rebased commit:

* `compile` on ubuntu-latest, macos-latest, and windows-latest
* `lint`
* `test-unit`
* `test-integration`
* `check_env`
* `license/cla`

The 21 E2E job failures are the fork-secrets limitation described above, not a code issue. GitHub does not pass repository secrets to workflows triggered from a fork, so the job logs show `Secret source: None` and an empty `PGP_PRIVATE_KEY`, which fails the gpg import step with exit code 2 before any CLI code runs. The deployer step then fails for the same reason once the image is pulled. A maintainer will need to trigger E2E from a branch in this repository to get real signal there.

This is ready for review whenever someone has capacity. Happy to split it into two PRs by command (`usermanagement` and `accessmanagement` separately) if that would make review easier.

PR #1844, opened 2026-05-13
fix: return non 0 with unsuported recipe

vjripoll:
**Description**
---
Fixes exit code behavior when targeted recipe installation fails.

**Problem**
---
When users specify a recipe explicitly using -n/--recipe (targeted install), if the installation fails (e.g., unsupported platform, recipe not found), the CLI incorrectly returns exit code 0 (success):

  newrelic install -n unsupported-recipe
  echo $?  # Returns 0 (wrong!)

This happens because the error handling falls through to the shouldInstallCore() block, which is designed for graceful fallback in guided installation mode. It prints a helpful message but doesn't propagate the error, causing the command to exit successfully.

**Solution**
---
Add an early return for targeted installs that propagates the error before reaching the fallback logic:

For targeted installs, return the error to ensure non-zero exit code
```
  if len(extractedRecipeNames) > 0 {
      return err
  }
```

  This ensures:
  - Targeted installs (-n recipe-name): Exit with non-zero code on failure ✅
  - Guided installs (no -n flag): Keep graceful fallback behavior ✅

---
BEFORE
---
Screenshot 2026-06-03 at 15 29 51

---
AFTER
---

Screenshot 2026-06-03 at 18 34 05


internal/install/command.go
```diff
@@ -116,6 +116,11 @@ var Command = &cobra.Command{
 				return err
 			}
 
+			// For targeted installs, return the error to ensure non-zero exit code
+			if len(extractedRecipeNames) > 0 {
+				return err
+			}
```
alvarocabanas, line 122: LGTM

alvarocabanas (approved):
lgtm

PR #1865, opened 2026-07-16
fix(install): exit non-zero when a targeted install installs no recipes

pranav-new-relic:
**Description**

When users specify recipes via `-n/--recipe` and none of the requested recipes end up installed (for example, every requested recipe is unsupported on the host OS), the CLI currently exits `0`. Automation wrapping the CLI cannot distinguish between "install succeeded" and "install did nothing," which masks failures during unattended provisioning. This change adds an `ErrNoRecipesInstalled` sentinel and returns a non-zero exit code specifically for the fully-failed targeted-install case.

A broader change — returning non-zero whenever any targeted recipe fails — was considered but rejected. That approach would flip the outcome of installs that partially succeed (for example, two of three requested recipes install cleanly while one is unsupported), and some existing pipelines legitimately treat a mixed outcome as an acceptable state. Scoping the fix to the "zero installed" case addresses the reported problem without changing behavior for partial-success runs, so the additional risk surface stays minimal.

### DIFF
--- a/internal/install/command.go
+++ b/internal/install/command.go
@@ -116,6 +116,10 @@ var Command = &cobra.Command{
 				return err
 			}
 
+			if len(extractedRecipeNames) > 0 && errors.Is(err, types.ErrNoRecipesInstalled) {
+				return err
+			}
+
 			if i.shouldInstallCore() {
 				fallbackErrorMsg := fmt.Sprintf("\nWe encountered an issue during the installation: %s.", err)
 				fallbackHelpMsg := "If this problem persists, visit the documentation and support page for additional help here at https://docs.newrelic.com/docs/infrastructure/install-infrastructure-agent/get-started/requirements-infrastructure-agent/"
--- a/internal/install/recipe_installer.go
+++ b/internal/install/recipe_installer.go
@@ -503,9 +503,7 @@ func (i *RecipeInstall) installAdditionalBundle(bundler RecipeBundler, bundleIns
 			}
 		}
 
-		return &types.UncaughtError{
-			Err: fmt.Errorf("no recipes were installed"),
-		}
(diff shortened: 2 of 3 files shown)


pranav-new-relic:
### Quick behavior reference

Concrete examples of how this change affects the CLI's exit code. Only the first two rows change; everything else behaves exactly as it does today.

| Command | Recipes installed | Exit code before | Exit code after |
|---|---|---|---|
| `newrelic install -n agent-control` *(agent-control unsupported on this OS)* | 0 | `0` (misleading) | `1` |
| `newrelic install -n a -n b` *(both unsupported)* | 0 | `0` (misleading) | `1` |
| `newrelic install -n a -n b -n c` *(only `b` unsupported)* | 2 | `0` | `0` *(unchanged)* |
| `newrelic install -n a` *(installs cleanly)* | 1 | `0` | `0` *(unchanged)* |
| `newrelic install` *(guided, no `-n`)* | any | `0` | `0` *(unchanged)* |

### Why this matters

Automation wrapping the CLI (CI steps, provisioning scripts, cloud-init) checks `$?` to decide whether the install succeeded. When the user explicitly targets one or more recipes with `-n` and none of them end up installed, exit `0` is a lie — the CLI reports "unsupported" on screen but tells shell-level tooling everything went fine. This change makes the exit code match what the user actually sees.

### Why not just fail on any partial failure

The partial-success case (row 3 above) is deliberately left at exit `0`. Some pipelines rely on that behavior when installing a set of recipes across a heterogeneous fleet where a few are expected to be unsupported. Flipping that outcome to non-zero would be a separate, larger discussion. Scoping this fix to "zero recipes installed" keeps the blast radius minimal while still fixing the reported bug.

nr-developer-toolkit (approved):

PR #1870, opened 2026-08-05
feat: new uninstall command

vjripoll:
**Summary**

Adds a new newrelic uninstall command that removes a previously installed integration/agent, following the same recipe-based model newrelic install uses.

Removal works exactly like installation: if the recipe defines its own Uninstall go-task Taskfile, the command runs it as-is, using the same go-task execution engine install already uses (GoTaskRecipeExecutor.ExecuteTaskfile). If a recipe defines no Uninstall taskfile, the command reports that no automated uninstall is available yet for it — it does not attempt to guess how to remove anything.

**Testing**

Verified manually on both Linux and Windows (see screenshots below). Added unit tests for the new internal/uninstall package (recipe finder, process detector, confirmer, recipe uninstaller, command) plus an integration test against the test-uninstall-taskfile fixture.

 LINUX
---
image

WINDOWS
---
image

### DIFF
--- a/cmd/newrelic/main.go
+++ b/cmd/newrelic/main.go
@@ -31,6 +31,7 @@ import (
 	"github.com/newrelic/newrelic-cli/internal/profile"
 	"github.com/newrelic/newrelic-cli/internal/reporting"
 	"github.com/newrelic/newrelic-cli/internal/synthetics"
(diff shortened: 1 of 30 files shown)


alvarocabanas:
We are closing this change since we opted for another path.

PR #1874, opened 2026-08-19
fix(fleetcontrol): paginate fleet search across all pages

ShaneLillieNR:
`fleetcontrol fleet search --name-equals`/`--name-contains` only inspected the first page of `GetEntitySearch` results, so the client-side name filter silently missed any fleet that wasn't on page 1 (confirmed live: a known fleet absent from a 100-entity first page, found instantly via a name-scoped query against the general `entitySearch` API).

`searchFleetEntities` now loops on `NextCursor` (capped at 1000 pages as a safety net against a misbehaving cursor) and accumulates entities across every page before applying the name filter. This intentionally does not expose a `--next-cursor` flag the way `fleet members list` does - a name search's entire point is "does this exist," and the caller has no way to know which page a match might be on, so partial/manual paging would defeat the feature.

Depended on newrelic/newrelic-client-go#1442, which fixed the same bug one layer down: `GetEntitySearch`'s cursor parameter was a no-op because the generated query never declared/passed `$cursor`. **That PR has merged and shipped in client-go v2.93.2** - `go.mod` now points at the real release instead of the temporary commit-pinned replace directive, so this is ready for review.

## Test plan

- `internal/fleetcontrol/fleet_management_search_test.go`: 5 unit tests against a mock NerdGraph server - match on page 1, match only on page 2+ (the regression test for this bug), no match across all pages, error on a later page propagates (no partial results), `--name-contains` spanning multiple pages.
- Verified the page-2 regression test isn't vacuous: temporarily capped the loop at one iteration and confirmed that test fails, then restored the fix.
- `make test-unit` - 526 tests passing. `gofmt` / `golangci-lint` clean.
- Live-verified against a real account (~800 fleets) with the actual built binary: unfiltered search returns all 798 fleets with no duplication, and `--name-equals` finds fleets at page 2+ that a pre-fix build returns `null` for. Details in the PR comments.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -76,3 +76,7 @@ require (
 )
 
 replace github.com/go-task/task/v3 => github.com/newrelic-forks/task/v3 v3.11.0
+
+// TEMPORARY: points at newrelic/newrelic-client-go#1442 (fix/fleetcontrol-entity-search-cursor)
+// for local testing while that PR is in review. Revert to a released version before merging.
+replace github.com/newrelic/newrelic-client-go/v2 => github.com/newrelic/newrelic-client-go/v2 v2.93.1-0.20260810175905-93cb87f7dc94
--- a/go.mod
+++ b/go.mod
@@ -17,7 +17,7 @@ require (
 	github.com/jedib0t/go-pretty/v6 v6.4.4
 	github.com/joshdk/go-junit v0.0.0-20210226021600-6145f504ca0d
 	github.com/mitchellh/go-homedir v1.1.0
-	github.com/newrelic/newrelic-client-go/v2 v2.91.0
+	github.com/newrelic/newrelic-client-go/v2 v2.93.2
 	github.com/pkg/errors v0.9.1
 	github.com/shirou/gopsutil/v3 v3.23.9
 	github.com/sirupsen/logrus v1.9.3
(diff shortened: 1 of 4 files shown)


ShaneLillieNR:
Live end-to-end verification of this fix, using the built `newrelic-cli` binary against a real account (~800 fleets), not just the unit tests in this PR:

**With this fix** (built against newrelic/newrelic-client-go#1442's commit via the temporary `go.mod` replace directive):
- Unfiltered `fleetcontrol fleet search` returns **798 fleets**, all 798 IDs unique — confirms the pagination loop walks every page (~8 pages at 100/page) with no overlap or duplication.
- `--name-equals` correctly finds fleets that only exist beyond page 1 — tested one at result index 150 and the very last one at index 797.

**Without this fix** (same live account, same two fleet names, built from `main` with unpatched `newrelic-client-go` v2.91.0 in a throwaway worktree): both `--name-equals` queries returned `null`. That's the exact bug being fixed — silently wrong empty results for any fleet not on page 1, no error, no warning.

pranav-new-relic (dismissed):
LGTM (as discussed over the call). Thanks for fixing this :)

pranav-new-relic (approved):

pranav-new-relic (approved):

PR #1886, opened 2026-09-09
fix(release): stop fetching install.ps1 from GitHub Release — use checkout copy instead

pranav-new-relic:
## Summary

- Removes the `dsaltares/fetch-gh-release-asset@0.06` step in `release-publish-windows.yml` that was downloading `install.ps1` from the GitHub Release asset before uploading it to S3
- The `actions/checkout@v3` step at the top of this job already provides `./scripts/install.ps1` — using it directly mirrors exactly how `release.yml` already handles `install.sh`
- Fixes `install.ps1` showing as 0 bytes at `https://download.newrelic.com/install/newrelic-cli/scripts/install.ps1`

## Confirmed prior art — how install.sh is already handled

In `.github/workflows/release.yml` at **line 176**, `install.sh` is uploaded to S3 directly from the checkout copy, with no intermediate fetch from the GitHub Release:

```yaml
# .github/workflows/release.yml — line 176
- name: Upload Unix based install script to AWS
  id: upload-install-script
  run: |
    aws s3 cp ./scripts/install.sh s3://nr-downloads-main/install/newrelic-cli/scripts/install.sh --profile virtuoso
```

No fetch step precedes it. The checkout is the source of truth. `install.sh` has worked reliably because of this. This fix makes `install.ps1` follow the identical pattern.

## Root cause

The broken step was overwriting the good local checkout file with a potentially 0-byte download:

```yaml
# This was overwriting ./scripts/install.ps1 with a 0-byte file on silent failure
- name: Fetch Github Release Asset - install.ps1
  uses: dsaltares/fetch-gh-release-asset@0.06   # very old action, silently exits 0 on failure
  with:
    file: "install.ps1"
    target: "./scripts/install.ps1"              # overwrites the checkout copy

# This then uploaded the 0-byte file to S3
- name: Upload Windows install script to AWS S3
  run: aws s3 cp ./scripts/install.ps1 s3://nr-downloads-main/install/newrelic-cli/scripts/install.ps1
```

## Test plan

- [ ] Confirm `install.ps1` at `s3://nr-downloads-main/install/newrelic-cli/scripts/install.ps1` is non-zero after the next release
- [ ] Confirm `NewRelicCLIInstaller.msi` upload (which still uses the fetch action legitimately — the MSI is not in the repo) is unaffected

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/.github/workflows/release-publish-windows.yml
+++ b/.github/workflows/release-publish-windows.yml
@@ -90,15 +90,6 @@ jobs:
           target: "NewRelicCLIInstaller.msi"
           token: ${{ secrets.RELEASE_TOKEN }}
 
-      - name: Fetch Github Release Asset - install.ps1
-        uses: dsaltares/fetch-gh-release-asset@0.06
-        with:
-          repo: "newrelic/newrelic-cli"
-          version: "tags/${{ steps.get-latest-tag.outputs.tag }}"
-          file: "install.ps1"
-          target: "./scripts/install.ps1"
-          token: ${{ secrets.RELEASE_TOKEN }}
-
       - name: Upload Windows install script to AWS S3
         id: upload-install-script-windows
         run: |


pranav-new-relic:
Hey @Nandu-pns — requesting your review here. Let me walk through exactly why this change is correct and why it can't break anything.

---

### What was happening (the bug)

The workflow that publishes `install.ps1` to S3 had **two jobs** in sequence:

1. `release.yml` → runs goreleaser → goreleaser attaches `scripts/install.ps1` from the repo as a GitHub Release asset
2. `release-publish-windows.yml` → supposed to upload `install.ps1` to S3

Inside job 2, the logic was:

```
checkout repo (ref: main)           ← ./scripts/install.ps1 is NOW on disk, correct, non-zero
        |
        v
fetch install.ps1 from GitHub Release  ← OVERWRITES ./scripts/install.ps1 with a download
        |
        v
aws s3 cp ./scripts/install.ps1 → S3  ← uploads whatever is at that path now
```

The `dsaltares/fetch-gh-release-asset@0.06` action is ancient (the repo is on v2+ now). At version `0.06`, when a fetch fails — network blip, permissions hiccup, timing issue — it **silently exits 0 and writes a 0-byte file**. It doesn't fail the step, so the workflow happily continues, uploads the 0-byte file to S3, and you end up with an empty `install.ps1` at the download URL. That's your bug.

---

### Why the fix is definitely correct

**The checkout copy is the authoritative source.** Here's the proof:

The goreleaser config (`.goreleaser.yml` lines 46–47) says:
```yaml
release:
  extra_files:
    - glob: scripts/install.ps1
      name_template: "install.ps1"
```

Goreleaser reads `scripts/install.ps1` from the repo and attaches it to the GitHub Release **unmodified** — no transformation, no compilation, just a file copy. So the GitHub Release asset and the file in `./scripts/install.ps1` after checkout are byte-for-byte identical by construction. Fetching from the release and then uploading is a redundant round-trip through GitHub's CDN with no benefit.

The workflow already has `actions/checkout@v3` with `ref: main` at the top. That gives us `./scripts/install.ps1` from the exact same commit that goreleaser built from. We don't need to fetch it again.

**This is already how `install.sh` works.** In `release.yml` line 176:
```yaml
aws s3 cp ./scripts/install.sh s3://nr-downloads-main/install/newrelic-cli/scripts/install.sh --profile virtuoso
```
No fetch step, no round-trip — just use the checkout copy directly. `install.sh` has never had this problem. The fix makes `install.ps1` follow the exact same pattern.

---

### Why this cannot break anything

Two concerns worth ruling out:

**1. "What if the checkout copy is somehow out of date?"**
The checkout uses `ref: main` and `fetch-depth: 0`. It's pulling the latest `main`. The goreleaser job in `release.yml` (which runs before this one) also checked out `main`. They're pulling the same ref. There is no version skew possible.

**2. "What if we still need the fetch step for some other reason?"**
The only thing the fetch step did was download `install.ps1` and write it to `./scripts/install.ps1`. Since that file already exists from checkout and the content is identical, removing the step changes nothing about what gets uploaded to S3 — except now we're uploading the copy that definitely has content instead of the one that might be 0 bytes.

The `NewRelicCLIInstaller.msi` fetch step is still there and untouched. That one is legitimately necessary — the MSI is built in a separate job (`release-windows-installer`) and attached to the GitHub Release. It's not in the repo checkout. That's a genuinely different case. The `install.ps1` case was never like that.

---

**TL;DR:** The deleted step was fetching a file we already had, using an old action that silently fails, and overwriting the good local copy with a potentially empty one. Removing it makes the workflow simpler and fixes the 0-byte S3 object.

Nandu-pns (approved):

